PR #2, opened 2018-02-05
New spectral edge target

chakshinglee:
The main goal of this PR is to add functionality of using eigen problem on local dual graph to generate edge traces. 

There are totally 5 options (including the original method), and some involves generalized eigen problem. So the (dense and sparse) solvers for generalized eigen problem are also added. The main modification for the additional trace generation methods is in `LocalMixedGraphSpectralTargets.xpp`.

The way we call `ARPACK` is modified, instead of doing the shift mode manually, the current implementation follow the the instruction on https://perso.univ-rennes1.fr/yvon.lafranche/mintera/index.html. 

Since there are 4 eigensolvers (dense, sparse, regular, generalized), so this PR also attempts to unify the interface (extract the dense Eigensolver from `utilities.xpp` to `LocalEigenSolver.xpp`). This also simplifies the `LocalMixedGraphSpectralTargets` where these eigensolvers are called.


examples/poweriter.cpp
```diff
@@ -41,6 +41,7 @@ int main(int argc, char* argv[])
     constexpr auto coarse_factor = 80;
     constexpr auto max_evects = 4;
     constexpr auto spect_tol = 1.0;
+    constexpr auto trace_method = 1.0;
```
barker29: Shouldn't this be an int? (ie 1 instead of 1.0)

examples/finitevolume.cpp
```diff
@@ -81,10 +81,12 @@ int main(int argc, char* argv[])
     bool hybridization = false;
     args.AddOption(&hybridization, "-hb", "--hybridization", "-no-hb",
                    "--no-hybridization", "Enable hybridization.");
+    int trace_method = 1;
+    args.AddOption(&trace_method, "-tm", "--trace-method",
+                   "Different methods (1-4) to get edge trace samples.");
```
barker29: Should be (1-5).

examples/generalgraph.cpp
```diff
@@ -104,6 +104,9 @@ int main(int argc, char* argv[])
     int isolate = -1;
     args.AddOption(&isolate, "--isolate", "--isolate",
                    "Isolate a single vertex (for debugging so far).");
+    int trace_method = 1;
+    args.AddOption(&trace_method, "-tm", "--trace-method",
+                   "Different methods (1-4) to get edge trace samples.");
```
barker29: Should be (1-5).

examples/timestep.cpp
```diff
@@ -106,6 +106,9 @@ int main(int argc, char* argv[])
     int k = 1;
     args.AddOption(&k, "-k", "--level",
                    "Level. Fine = 0, Coarse = 1");
+    int trace_method = 1;
+    args.AddOption(&trace_method, "-tm", "--trace-method",
+                   "Different methods (1-4) to get edge trace samples.");
```
barker29: (1-5), also, is there a way to briefly identify the methods in this help message?

src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -109,13 +110,22 @@ class LocalMixedGraphSpectralTargets
...
+    void CheckMinimalEigenvalue(
+        double eval_min, int aggregate_id, std::string entity);
+
     MPI_Comm comm_;
 
-    double rel_tol_;
-    int max_evects_;
+    const double rel_tol_;
+    const int max_evects_;
+    const int trace_method_;
```
barker29: I think this should be an enumeration, so code maintainers have at least some idea what the methods are.

src/LocalMixedGraphSpectralTargets.cpp
```diff
@@ -22,7 +22,7 @@
 #include "GraphCoarsen.hpp"
 #include "utilities.hpp"
 #include "sharedentitycommunication.hpp"
-#include "arpackeig.hpp"
```
barker29, line 25: I think we should entirely remove `arpackeig.?pp` in this PR.

src/LocalMixedGraphSpectralTargets.cpp
```diff
@@ -223,11 +292,7 @@ void LocalMixedGraphSpectralTargets::ComputeVertexTargets(
     // ---
     // solve eigenvalue problem on each (extended) neighbor_agg, our (3.1)
     // ---
-    Eigensolver eigs;
-    bool use_dense_eigensolver = true;
-#if SMOOTHG_USE_ARPACK
-    SparseEigensolver speigs;
-#endif
+    LocalEigenSolver eigs(max_evects_, rel_tol_);
```
gelever, line 297: This looks much cleaner now 👍

src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -62,13 +63,13 @@ class LocalMixedGraphSpectralTargets
        \f]
     */
     LocalMixedGraphSpectralTargets(
-        double rel_tol, int max_evects,
+        double rel_tol, int max_evects, int trace_method,
```
gelever: A named enum would be more clear than an int, but it would be difficult to name them something short and succinct  `TraceType::Normal` `TraceType::Dual`, etc

src/MatrixUtilities.hpp
```diff
@@ -111,11 +111,32 @@ mfem::SparseMatrix SparseIdentity(int size);
    @brief Construct an rectangular identity matrix (as a SparseMatrix)
    @param rows number of row
    @param cols number of columns
-   @params row_offset offset row where diagonal identity starts
-   @params col_offset offset column where diagonal identity starts
+   @param row_offset offset row where diagonal identity starts
```
gelever, line 114: 👍

testcode/lineargraph.cpp
```diff
@@ -220,11 +220,11 @@ int main(int argc, char* argv[])
...
                                  *partition.face_d_td_d);
 
-    std::vector<mfem::DenseMatrix> local_edge_traces(num_partitions - 1);
-    std::vector<mfem::DenseMatrix> local_spectral_vertex_targets(num_partitions);
+    std::vector<mfem::DenseMatrix> local_edge_traces;
+    std::vector<mfem::DenseMatrix> local_spectral_vertex_targets;
 
     LocalMixedGraphSpectralTargets localtargets(
-        spect_tol, max_evects, graph.GetM(), graph.GetD(), graph_topology);
+        spect_tol, max_evects, 1, graph.GetM(), graph.GetD(), graph_topology);
```
gelever: I think it would be more clear to name the variable so that it isn't necessary to look up what `1` does.

```c++
int trace_method = 1;
LocalMixedGraphSpectralTargets localtargets(spect_tol, max_evects, trace_method, graph.GetM(), graph.GetD(), graph_topology);
```

gelever (approved):
Looks good @chakshinglee, much better handling of the eigensolvers.

src/LocalMixedGraphSpectralTargets.hpp
chakshinglee: Agreed (both first and second part). Maybe it's easier to give them more concrete names after we decide which ones to keep.
- [ ] Use a named enum for `trace_method`

src/SpectralAMG_MGL_Coarsener.hpp
```diff
@@ -46,13 +46,15 @@ class SpectralAMG_MGL_Coarsener : public Mixed_GL_Coarsener
               accurate, more expensive coarse space.
        @param max_evecs_per_agg cap the number of eigenvectors per aggregate that will
               be used in the coarse space.
+       @param trace_method methods (1-4) for getting edge trace samples
```
barker29: (1-5).

src/LocalMixedGraphSpectralTargets.hpp
chakshinglee: @gelever How should we deal with command line option parsing for enum type? Convert manually?

barker29 (approved):
This is a very nice addition.

I think we can remove `arpackeig.?pp` entirely from the repo now.

With regard to the different methods: a named enum would be better than the numbers (as discussed), but I propose another possibility:

Separate the new methods into several choices/options, for example `dual_targets` uses the dual graph Laplacian (with no other options this is M2), then `scaled_dual` and `energy_dual` can be used in combination to make the other possibilities. This has the advantage that we can remember what they do and their tradeoffs (ie, document that `scaled_dual` improves error but costs in iterations, while `energy_dual` gives a nice theoretical guarantee for the approximation property). The disadvantage is now we have three options instead of one...

src/LocalMixedGraphSpectralTargets.hpp
gelever: The simplest way would be to parse an int and then static cast to the enum.

src/LocalMixedGraphSpectralTargets.cpp
chakshinglee: good catch, I forgot to do that.

examples/timestep.cpp
chakshinglee: You suggestion above is more logical than listing methods 1-5. It is not too bad to me to have 2 more parameters. Maybe we can make a `UpscaleParameters` struct later to store all these parameters? I think I have to do this (make a parameter struct) for the other branch when `saamge` is called.

PR #5, opened 2018-02-06
Feature/travis ci

gelever:
I am now content w/ this.  Builds, compiles, runs test (including style, valgrind).  The examples are not tested because they take forever to run, but it would not be difficult to add them again.

~This is as far I got.  The dependencies compile and are cached.  So if for some reason they need to be rebuilt, it should be sufficient to delete the install directory.~

~smoothG also compiles.  However, the tests don't do so hot.  Mpich has something going on with executable paths (?).  So the parallel tests don't run.  I couldn't get it to build w/ openmpi.  And openmpi doesn't play nice w/ valgrind.~

- [x] ~fix tests~
- [x] ~parallel tests are super slow~

~using both mpich and openmpi seems to have done something: https://travis-ci.org/LLNL/smoothG/builds/337806566, at least parallel tests run doesn't need both for tests to pass, runs fine w/ mpich for now, but valgrind still broken~


barker29:
This is a lot of work, @gelever , thank you. For now we can skip the parallel tests and/or the valgrind tests if that makes it easier - I will be happy to just have a CI system that builds and does very basic tests for the moment.

gelever:
@barker29 ok.  I'll try to build mpich to see if it helps with valgrind.  Then we can run some limited number of tests, maybe some sort of `make check`

smoothg.astylerc
```diff
@@ -33,7 +33,7 @@
 --formatted
 #--break-blocks
 --pad-oper
---pad-comma
+#--pad-comma
```
gelever: This doesn't exist in Astyle version 2 which is what is in the ubuntu repo for travis-ci

barker29 (approved):
Looks good! Thanks for doing this.

chakshinglee (approved):
Looks great, thanks!

PR #6, opened 2018-02-06
Fix/housekeeping

gelever:
Separates README.md from doxygen front page.   This way the main github page looks better without all the doxygen markup.  It will probably be a bit more work to keep them in sync, but the trade off is worth it.

### DIFF
--- a/Doxyfile.in
+++ b/Doxyfile.in
@@ -889,7 +889,7 @@ FILTER_SOURCE_PATTERNS =
 # (index.html). This can be useful if you have a project on for instance GitHub
 # and want to reuse the introduction page also for the doxygen output.
 
-USE_MDFILE_AS_MAINPAGE =
+USE_MDFILE_AS_MAINPAGE = @PROJECT_SOURCE_DIR@/doc/front.md
 
 #---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


gelever:
`README.md`'s example link should now probably point to the `examples` dir.  `doc/EXAMPLE.md` does not much sense without `latex` and code snippets.  Or create a separate annoted `EXAMPLE.md` that works in markdown.

barker29 (approved):
This is probably an improvement but I don't love it. In general I am not happy with how doxygen interfaces with github, but I'm not sure how to fix it.

gelever:
I would prefer github look nice over doxygen.  Keeping the two separate maybe a better solution.  Or maybe use doxygen only for code documentation and the rest in normal markdown so that it looks nice on github.

barker29:
I agree with you, but I keep hoping there's a way to make them *both* look nice, without duplicating things.

gelever:
I would not rely too much on doxygen and suggest to limit it to only api documentation. Then implement the guides in regular markdown that doxygen does not reference at all.  For example, [RAJA](https://github.com/LLNL/RAJA) has both clear documentation and doxygen support.

PR #7, opened 2018-02-08
Saamge

chakshinglee:
The main goal of this PR is to add option to call [saamge_parallel (smoothed aggregation spectral element-based AMG)](https://github.com/LLNL/saamge) in the hybridization solver. Simple usage of this feature is illustrated in `graphupscale.cpp`. 

There is another additional feature in `HybridSolver`, which is to rescale the hybridized system by a diagonal scaling vector (computed by a few smoothing steps of the original hybridized system). In some cases, this may improve the performance of the solver (not guarantee). This offers a simpler alternative to improve solver performance without adding extra dependencies.


chakshinglee:
The rescaling is described in [this article](https://link.springer.com/chapter/10.1007/978-3-319-52389-7_6) (around middle of page 6).

src/FiniteVolumeUpscale.cpp
```diff
@@ -33,7 +33,8 @@ FiniteVolumeUpscale::FiniteVolumeUpscale(MPI_Comm comm,
                                          const mfem::Array<int>& ess_attr,
                                          double spect_tol, int max_evects,
                                          bool dual_target, bool scaled_dual,
-                                         bool energy_dual, bool hybridization)
+                                         bool energy_dual, bool hybridization,
+                                         SAAMGeParam* saamge_param)
```
gelever: This can probably be const (?)

src/FiniteVolumeUpscale.cpp
```diff
@@ -69,9 +70,19 @@ FiniteVolumeUpscale::FiniteVolumeUpscale(MPI_Comm comm,
 
     if (hybridization) // Hybridization solver
     {
-        coarse_solver_ = make_unique<HybridSolver>(
-                             comm, mixed_laplacians_.back(), *coarsener_,
-                             &coarsener_->get_GraphTopology_ref().face_bdratt_, &marker);
+        auto face_bdratt = coarsener_->get_GraphTopology_ref().face_bdratt_;
+        if (saamge_param)
```
gelever: This would probably be cleaner to just pass the pointer directly to the solver and check if it is null there.  But there are already a lot of constructor parameters for `HybridSolver`. I'm not sure that it would be easy to arrange them in a reasonable manner that would satisfy all scenarios.

src/HybridSolver.cpp
```diff
@@ -799,6 +849,90 @@ void HybridSolver::RecoverOriginalSolution(const mfem::Vector& HybridSol,
...
+    cg_scale.SetPreconditioner(prec_scale);
+    cg_scale.SetOperator(*tmpH);
+    cg_scale.Mult(zeros, diagonal_scaling_);
+
+    auto Scale = VectorToMatrix(diagonal_scaling_);
+    mfem::HypreParMatrix pScale(comm_, tmpH->N(), tmpH->ColPart(), &Scale);
+    pHybridSystem_.reset(smoothg::RAP(*tmpH, pScale));
+}
+
+void HybridSolver::RescaleVector(mfem::Vector& vec) const
```
gelever: If this would be useful elsewhere, it might fit better in `utilities.xpp`

examples/graphupscale.cpp
```diff
@@ -136,33 +136,50 @@ int main(int argc, char* argv[])
...
 
+#if SMOOTHG_USE_SAAMGE
+        SAAMGeParam saamge_param;
+        const auto hbsa_upscale = GraphUpscale(comm, vertex_edge, coarse_factor,
+                                               spect_tol, max_evects, dual_target,
+                                               scaled_dual, energy_dual, use_hybridization,
+                                               mfem::Vector(), &saamge_param);
+
+        const auto hbsa_sol = hbsa_upscale.Solve(rhs_u_fine);
+        const auto error_sa = CompareError(comm, hbsa_sol, minres_sol);
```
gelever: `HB (SAAMGe) vs HB (BoomerAMG) Error` would be interesting to see too.

examples/finitevolume.cpp
```diff
@@ -173,17 +173,9 @@ int main(int argc, char* argv[])
...
-    {
-        mfem::SparseMatrix tmp = TableToSparse(pmesh->ElementToEdgeTable());
-        vertex_edge.Swap(tmp);
-    }
-    else
-    {
-        mfem::SparseMatrix tmp = TableToSparse(pmesh->ElementToFaceTable());
-        vertex_edge.Swap(tmp);
-    }
+    auto& vertex_edge_table = nDimensions == 2 ? pmesh->ElementToEdgeTable()
```
gelever, line 176: 👍

src/HybridSolver.hpp
```diff
@@ -90,36 +112,59 @@ class HybridSolver : public MixedLaplacianSolver
...
+
+    /**
+       @brief Constructor for coarse-level hybridiziation solver.
+
+       SAAMGe is used as preconditioner.
+
+       @param comm MPI communicator
+       @param mgL Mixed matrices for the graph Laplacian in the coarse level
+       @param mgLc Mixed graph Laplacian Coarsener from fine to coarse level
+       @param saamge_param SAAMGe paramters
```
gelever: `paramters` -> `parameters`
:D

gelever (approved):
Looks good @chakshinglee. 

All of the setup parameters are starting to get complex, so we might want to eventually figure out a way to simplify that.

src/HybridSolver.hpp
chakshinglee: Good catch!

barker29:
Are you using SAAMGe from institutional bitbucket? Does the version/hash matter?

chakshinglee:
My bad. Yes, I am using the SAAMGe from institutional bitbucket, and I think the current master should be ok (or versions later than 5b8a9e1f8ce). Essentially we need `ElementMatrixDenseArray`.

CMakeLists.txt
```diff
@@ -44,6 +44,26 @@ set(TPL_LIBRARIES "")
 # This is a list of linker flags to be used with TPLs for all targets
 set(TPL_LINKER_FLAGS "")
 
+# SAAMGE_PARALLEL
```
chakshinglee, line 47: @barker29 @gelever Is the order of adding TPL important? I had no problem of compiling smoothG on a Macbook (before this commit), but on my tux box I cannot compile some of the examples (undefined reference to ~socketstream). If I add saamge prior to mfem, the linking errors are gone (note saamge depends on mfem as well). But I am not sure if this is the actual problem (or I just build saamge in a wrong way).

gelever: As far as I can tell, the order does matter.  It should  be in descending order, where thing up top depend on things below them.  Since saamge depends on mfem, it should be placed before mfem.

src/FiniteVolumeUpscale.cpp
chakshinglee: This is a good point.

CMakeLists.txt
chakshinglee: I see, then I will keep this commit. Thanks!

src/MatrixUtilities.hpp
```diff
@@ -41,19 +41,22 @@ void Print(const mfem::SparseMatrix& mat, const std::string& label = "",
...
 /**
     @brief Creates a sparse matrix from a table
 */
-mfem::SparseMatrix TableToSparse(const mfem::Table& table);
+mfem::SparseMatrix TableToMatrix(const mfem::Table& table);
+
+/**
+    @brief Creates a table from a sparse matrix's graph
+*/
+mfem::Table* MatrixToTable(const mfem::SparseMatrix& mat);
```
gelever: Is this change an issue with `Table` or some change for `SAAMGE` compatibility?

src/HybridSolver.cpp
```diff
@@ -816,48 +821,48 @@ void HybridSolver::BuildSpectralAMGePreconditioner()
...
-    nparts_arr[0] = (num_elems / saamge_param_->first_coarsen_factor) + 1;
+    int num_elems = elem_elem->Size();
+    sa_nparts_.resize(saamge_param_->num_levels - 1);
+    sa_nparts_[0] = (num_elems / saamge_param_->first_coarsen_factor) + 1;
 
     bool first_do_aggregates = (saamge_param_->num_levels <= 2 && saamge_param_->do_aggregates);
-    auto apr = saamge::agg_create_partitioning_fine(
-                   *pHybridSystem_, num_elems, &elem_dof, &elem_elem, nullptr, bdr_dofs.data(),
-                   &(nparts_arr[0]), multiplier_d_td_.get(), first_do_aggregates);
+    sa_apr_ = saamge::agg_create_partitioning_fine(
```
chakshinglee: @gelever Regarding your question, `sa_apr_` (pointer to `saamge::agg_partitioning_relations_t`) is constructed by with `elem_dof` and `elem_elem` (pointers to `mfem::Table`) as part of the input, and I think `sa_apr_` is supposed to own these table pointers because `agg_free_partitioning` (pretty much the destructor of `sa_apr_`) will `delete` those `mfem::Table` pointers. That's why I changed the return type. 

Another way to resolve problem is to create new pointers to those tables in `BuildSpectralAMGePreconditioner`, and `Swap` with the object returned by `MatrixToTable`. This way we can keep the origin return type of the function. What do you think about this approach?

gelever: I would avoid returning raw pointers if possible, especially just to satisfy an optional dependency.  Maybe just change to
 `mfem::Table* elem_dof = new Table(MatrixToTable(Agg_multiplier_));` or if that forces a copy, the swap idea works better.

chakshinglee: Got you, I will switch the return type back the previous setting. I think using swap is better (it's not as neat, but I do want to avoid copy one more time).

src/MetisGraphPartitioner.cpp
```diff
@@ -149,7 +161,7 @@ void MetisGraphPartitioner::SetPostIsolateVertices(const mfem::Array<int>& indic
 }
 
 mfem::SparseMatrix MetisGraphPartitioner::getAdjacency(
-    const mfem::SparseMatrix& wtable) const
+    const mfem::SparseMatrix& wtable)
```
barker29, line 164: Why remove the `const`?

barker29 (approved):
This looks good. Can we add a test for `GraphUpscale.cpp` so that this actually gets called in the test suite?

Also, just as a note to myself, SAAMGe produces a lot of output and needs a better interface for controlling that output.

src/MetisGraphPartitioner.cpp
gelever: The function is now static

chakshinglee:
If we want to add tests for showing the solvers work as expected, how about modifying `tinygraphsolver.cpp` to include different solvers? `graphupscale` seems to be a demonstration of what the `Upscale` class can do, a test for it may be just to run till exit with success?

barker29:
Sure, that makes sense. I just suggested `graphupscale` because that's the only place the saamge stuff appears in this PR.

src/MetisGraphPartitioner.cpp
chakshinglee: Yea, maybe this function should be taken out as an independent function (something like `RemoveDiagonal` in `MatrixUtilities.xpp`)?

testcode/tinygraphsolver.cpp
```diff
@@ -133,22 +137,21 @@ int main(int argc, char* argv[])
...
     mfem::BlockVector rhs = *mixed_graph_laplacian.subvecs_to_blockvector(rhs_u_fine, rhs_p_fine);
     mfem::BlockVector sol(blockOffsets);
-    sol = 0.0;
 
-    // solve
-    MinresBlockSolver mgp(comm, mixed_graph_laplacian);
-    mgp.SetPrintLevel(1);
-    mgp.Mult(rhs, sol);
+    // setup solvers
+    std::map<const MixedLaplacianSolver*, std::string> solver_to_name;
```
chakshinglee: @barker29 @gelever Here I am using the setting in `poweriter.cpp`. this file should contain the main new modifications, can you guys please take another look. After that I guess I will merge this branch.

testcode/tinygraphsolver.cpp
```diff
@@ -189,15 +192,45 @@ int main(int argc, char* argv[])
...
+
+        std::cout << "\nSolving mixed graph Laplacian problem with "
+                  << name << " ..." << std::endl;
 
-    truesol -= sol.GetBlock(1);
-    double norm = truesol.Norml2();
-    std::cout << "Error norm: " << norm << std::endl;
+        // solve
+        sol = 0.0;
+        const_cast<MixedLaplacianSolver*>(solver)->SetPrintLevel(1);
```
gelever: Avoiding this `const_cast` can be done by not having `const` in the map or setting the print level before placing into map.

 Otherwise, this looks good :+1:

testcode/tinygraphsolver.cpp
barker29: Looks good to me.

PR #8, opened 2018-02-12
Refactor GraphCoarsen::BuildPEdges() [feature/coarse mass]

barker29:
As I started to implement a new feature involving coarse mass matrices and `GraphCoarsen::BuildPEdges` I decided that method was too long and needed refactoring. This PR is the result.

The method is now shorter, and I think it is more readable, but this came at the cost of inventing a couple new objects and some new complexity.

I am particularly interested in @chakshinglee input on the names/etc of the new methods so they describe what they actually do.


src/GraphCoarsenBuilder.hpp
```diff
@@ -0,0 +1,179 @@
...
+class CoarseMBuilder
+{
+public:
+    virtual ~CoarseMBuilder() {}
+
+    /// The names of the next several methods are not that descriptive or
+    /// informative; they result from removing lines from BuildPEdges()
+    /// and putting it here.
+    virtual void RegisterRow(int agg_index, int row, int cdof_loc, int bubble_counter) = 0;
+
```
chakshinglee, line 54: This function is for setting the values in the (TraceExtension, Bubble)-block (and the transpose) of coarse M. How about call it `SetTraceBubbleBlock`?

src/GraphCoarsenBuilder.hpp
```diff
@@ -0,0 +1,179 @@
...
+    virtual void SetBubbleOffd(int l, double value) = 0;
+
+    virtual void AddDiag(double value) = 0;
```
chakshinglee: `AddDiag` and `AddTrace` are for setting the values in the (TraceExtension, TraceExtension)-block of coarse M, I think it's ok to use the same function name for these functions. For the name, follow similar logic `AddTraceBlock` or `AddTraceTraceBlock` .

src/GraphCoarsenBuilder.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    virtual void AddTrace(int l, double value) = 0;
+
+    virtual void SetBubbleLocal(int l, int j, double value) = 0;
```
chakshinglee: Similarly, `SetBubbleBlock` or `SetBubbleBubbleBlock`.

src/GraphCoarsenBuilder.cpp
```diff
@@ -0,0 +1,253 @@
...
+    edge_cdof_marker_ = -1;
+    edge_cdof_marker2_.SetSize(size);
+    edge_cdof_marker2_ = -1;
+}
+
+void AssembleMBuilder::ResetEdgeCdofMarkers(int size)
+{
+}
+
+void ElementMBuilder::RegisterTraceFace(int face_num, const mfem::SparseMatrix& face_Agg,
```
chakshinglee: `FillEdgeCdofMarkers` or `RegisterEdgeCdof` or `RegisterEdgeCdofMarkers`?

src/GraphCoarsenBuilder.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+    virtual void ResetEdgeCdofMarkers(int size) = 0;
+
+    virtual void RegisterTraceFace(int face_num, const mfem::SparseMatrix& face_Agg,
+                                   const mfem::SparseMatrix& Agg_cdof_edge) = 0;
+
+    /// Deal with shared dofs for trace
+    virtual void AddTraceAcross(int row, int col, double value) = 0;
```
chakshinglee: `AddDiag`, `AddTrace` and this function belong to the same group. 

They add contribution to (TraceExtension, TraceExtension)-block of coarse M.

src/GraphCoarsen.hpp
```diff
@@ -179,14 +181,63 @@ class GraphCoarsen
...
+                             const mfem::SparseMatrix& face_edge,
+                             const mfem::SparseMatrix& Agg_face);
+
+    /**
+       @brief takes the column column of potentials, multiplies
+       DtransferT by it, and returns the inner product with trace
+
+       helper for BuildPEdges
+    */
+    double DTTraceProduct(const mfem::SparseMatrix& DtransferT,
```
chakshinglee: I think this is a good name for what the function is doing. However, it may be hard to see why we want to compute this product (assuming the person only reads our paper). And apparently it is not easy to explain briefly. I will write a short note to explain this. You may see if anything we can do to make future developers/maintainers' lives easier.

chakshinglee (approved):
@barker29 the new objects/classes for building coarse M certainly make `BuildPEdges()` more readable. Thanks for cleaning up this ugly function.

gelever (approved):
Looks good @barker29.  Do you think it would be beneficial to move some of the shared stuff to the base class or would it be better to keep things separate?

barker29:
Good idea.

PR #9, opened 2018-02-15
Set pv trace when only 1 dofs on face 

chakshinglee:
add one more condition in several if statements so that when there is only 1 dof on a coarse face, the value of pv trace will be set to 1 instead of doing unnecessary computation. This kind of situation typically won't happen in graphs from meshes, but can happen in e.g. social networks.

### DIFF
--- a/src/LocalMixedGraphSpectralTargets.cpp
+++ b/src/LocalMixedGraphSpectralTargets.cpp
@@ -485,7 +485,7 @@ void LocalMixedGraphSpectralTargets::ComputeEdgeTargets(
 
         // restrict local sigmas in AggExt_sigma to the coarse face.
         // shared faces or interior faces
-        if (face_IsShared.RowSize(iface) || num_neighbor_aggs == 2)
+        if ((face_IsShared.RowSize(iface) || num_neighbor_aggs == 2) && num_iface_edge_dof > 1)
         {
             int total_vects = 0;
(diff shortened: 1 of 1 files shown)


barker29 (approved):
Looks fine to me!

src/LocalMixedGraphSpectralTargets.cpp
```diff
@@ -581,9 +581,9 @@ void LocalMixedGraphSpectralTargets::ComputeEdgeTargets(
                                              face_nbh_dofs, colMapper);
             sec_D.ReduceSend(iface, Dloc);
         }
-        else // global boundary face
+        else // global boundary face or only 1 dof on face
         {
-            mfem::SparseMatrix empty_matrix;
```
gelever, line 586: Did the difference between these `empty_matrix` and `SparseIdentity(0)` cause issues?  Just curious.

gelever (approved):
Looks good.

src/LocalMixedGraphSpectralTargets.cpp
chakshinglee: Yes. I think the problem is when we send/receive the matrix in SharedEntityCommunication, it will check the nnz (to know how many data to send), which is the last entry of I. But if we initialize a SparseMatrix without doing anything, I points to NULL.

PR #10, opened 2018-02-17
extract the process of building 'entity_trueEntity' 

chakshinglee:
extract the process of building 'entity_trueEntity' from 'entity_trueEntity_entity' to reduce repeated code.

(the block of code in `GraphCoarsen.cpp` is not exactly the same, but the results should be the same as it is based on the same logic when choose true entities)


gelever (approved):
Looks good @chakshinglee,  simplifies things nicely.

barker29 (approved):
Looks very nice.

PR #11, opened 2018-02-20
Fix/memory leak

chakshinglee:
Fix still reachable memory leak in `finitevolume.cpp` (the allocated memory in `InversePermeabilityFunction` is not deleted).

 There is also the `Conditional jump or move depends on uninitialised value(s)` error in the use of `arpackpp`. But I dont know how to fix that. It is not a leak?

### DIFF
--- a/examples/finitevolume.cpp
+++ b/examples/finitevolume.cpp
@@ -49,7 +49,7 @@ int main(int argc, char* argv[])
     picojson::object serialize;
 
     // 1. Initialize MPI
-    MPI_Init(&argc, &argv);
+    mpi_session session(argc, argv);
     MPI_Comm comm = MPI_COMM_WORLD;
     MPI_Comm_size(comm, &num_procs);
     MPI_Comm_rank(comm, &myid);
--- a/examples/spe10.hpp
+++ b/examples/spe10.hpp
@@ -247,6 +247,7 @@ SPE10Problem::SPE10Problem(const char* permFile, int nDimensions,
 
 SPE10Problem::~SPE10Problem()
 {
+    smoothg::InversePermeabilityFunction::ClearMemory();
     delete source_coeff_;
     delete kinv_;
     delete pmesh_;
--- a/examples/timestep.cpp
+++ b/examples/timestep.cpp
@@ -56,7 +56,7 @@ int main(int argc, char* argv[])
     picojson::object serialize;
 
     // 1. Initialize MPI
-    MPI_Init(&argc, &argv);
+    mpi_session session(argc, argv);
     MPI_Comm comm = MPI_COMM_WORLD;
     MPI_Comm_size(comm, &num_procs);
     MPI_Comm_rank(comm, &myid);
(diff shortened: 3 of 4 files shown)


chakshinglee:
The `Conditional jump or move depends on uninitialised value(s)` error is from [here](https://github.com/LLNL/smoothG/blob/7e2c7b426b6e7db52a44cd3d9c417e98a2f0dfe9/src/LocalEigenSolver.cpp#L426) (the same issue should be in the generalized eigen problem solver too). It seems to be some problem in `arpackpp` or I am not using it in a correct way. 

If instead of `Eigenvalues`, `EigenValVectors` is called to find the maximum eigenvalue, the error is gone. But I don't want to use the latter because the latter will compute the unnecessary eigenvector as well.

gelever:
@chakshinglee I think that conditional jump is a bug in `arpackpp`.  It doesn't set `HowMny` before calling the fortran routines for eigenvalues.  You can verify modifying `arrseig.h` at around line 1077 to
```cpp
template<class ARFLOAT, class ARTYPE>
int ARrcStdEig<ARFLOAT, ARTYPE>::FindEigenvalues()
{

  // Determining eigenvalues if they are not available.

  if (!ValuesOK) {
    try {
      ValAllocate();
      nconv = FindArnoldiBasis();
      rvec  = false;
      HowMny = 'A';   // <---- Add this line
      if (nconv>0) {
        Eupp();
        EuppError();
      }    
    }    
    catch (ArpackError) {
      ArpackError(ArpackError::CANNOT_FIND_VALUES, "FindEigenvalues");
      return 0;
    }    
    if (newVal) ValuesOK = true;
  }
  return nconv;

} // FindEigenvalues.
```
Edit: or just give it a default value in the constructor.  As long as it is given a value somewhere.


The fastest way to clear up the error might be to derive our own class, instead of modifying arpack directly:
```cpp
template<class ARFLOAT, class ARFOP>
class ARSymStdEig_: public virtual ARSymStdEig<ARFLOAT, ARFOP>
{
    public:
        using ARSymStdEig<ARFLOAT, ARFOP>::HowMny;

        ARSymStdEig_(int np, int nevp, ARFOP* objOPp,
                void (ARFOP::* MultOPxp)(ARFLOAT[], ARFLOAT[]),
                const std::string& whichp = "LM", int ncvp = 0, ARFLOAT tolp = 0.0,
                int maxitp = 0, ARFLOAT* residp = NULL, bool ishiftp = true)
            : ARSymStdEig<ARFLOAT, ARFOP>(np, nevp, objOPp, MultOPxp,
                    whichp, ncvp, tolp, maxitp, residp, ishiftp)
        {
            HowMny = 'A';
        }

        virtual ~ARSymStdEig_() { }

}; // class ARSymStdEig_.
```

gelever (approved):

barker29:
Can one of you report this issue at github.com/m-reuter/arpackpp ? It looks like there is some effort to maintain this code so it may be worth reporting.

chakshinglee:
@gelever thanks for the suggestion, I agree that's probably the best way to fix the issue for now. Do you want to make a commit to add the derived class (and use it in the corresponding places). And as @barker29 suggested, maybe @gelever you can report this to the arpackpp github site?

barker29 (approved):

gelever:
Issue at [https://github.com/m-reuter/arpackpp/issues/14](https://github.com/m-reuter/arpackpp/issues/14)

PR #12, opened 2018-02-22
Multilevel graph topology

chakshinglee:
This PR adds the feature of constructing multilevel graph topology. The idea should be straight forward (since originally we have a coarse level graph in a two level setting, which is `Agg_face`, and we just coarsen the coarse graph one more time, and the process continues until the number of requested levels is reached).

The way I do it is to add a constructor that takes a `GraphTopology` object as input. I think an alternative way is to make a `Coarsen` member function and return the coarsened `GraphTopology` object. We can  do the latter if we think that's clearer.


testcode/mltopo.cpp
```diff
@@ -0,0 +1,180 @@
...
+ * availability, see https://www.github.com/llnl/smoothG.
+ *
+ * smoothG is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License (as published by the Free
+ * Software Foundation) version 2.1 dated February 1999.
+ *
+ ***********************************************************************EHEADER*/
+
+/**
+   @file finitevolume.cpp
```
gelever: documentation is still from `finitevolume` :D

testcode/mltopo.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "../examples/spe10.hpp"
+
+#include "../src/picojson.h"
+#include "../src/smoothG.hpp"
+
+using namespace smoothg;
+
+void ShowAggregates(std::vector<GraphTopology>& graph_topos, mfem::ParMesh* pmesh);
+
+int main(int argc, char* argv[])
```
gelever: This test is pretty cool visualization tool. :+1:   Would this fit better as example or testcode?

src/GraphTopology.cpp
```diff
@@ -444,4 +481,54 @@ GraphTopology::GraphTopology(const mfem::SparseMatrix& face_edge,
...
+
+std::vector<GraphTopology> MultilevelGraphTopology(
+    mfem::SparseMatrix& vertex_edge, const mfem::HypreParMatrix& edge_d_td,
+    const mfem::SparseMatrix* edge_boundaryattr, int num_levels, int coarsening_factor)
+{
+    std::vector<GraphTopology> graph_topologies;
+    graph_topologies.reserve(num_levels - 1);
+
+    // Construct finest level graph topology
+    mfem::Array<int> partitioning;
```
gelever, line 509: Partitioning `vertex_vertex` is repeated a few times throughout the code/examples.  Maybe we should make a `PartitionAAT(vertex_edge, partitioning, coarsening_factor` utility function.  Not a huge deal tho, its only a few lines of code.

testcode/mltopo.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+   mpirun -n 4 ./finitevolume
+*/
+
+#include <fstream>
+#include <sstream>
+#include <mpi.h>
+
+#include "mfem.hpp"
+#include "../examples/spe10.hpp"
```
gelever: is anything from `spe10.hpp` used here?

gelever (approved):

testcode/mltopo.cpp
chakshinglee: good catch, I will fix it

testcode/mltopo.cpp
chakshinglee: good catch, I used it before, but not now.

testcode/mltopo.cpp
chakshinglee: Actually this file is more like an example (even the tests I set for it are just to run till exit). So I will move it to examples.

barker29 (approved):
I really like the visualization in `mltopo`.

PR #13, opened 2018-02-28
Allow some unbalance in processor partitioning

gelever:
Partitioning the mesh across processors outside MFEM lets us set an unbalance tolerance.
In my testing this produces better results as METIS has more play to optimize the number of edge cuts.

### On a once refined SPE10 mesh : 
Original, no unbalanced tolerance
```
time mpirun -np 8 ./finitevolume --dim 3 -m 4 -t .10 -ma -hb
Options used:
   --max-evects 4
   --spect-tol 0.1
   --metis-agglomeration
   --spe10-scale 5
   --hybridization
// ...  irrelevant args removed to save space
3438088 fine edges, 3401664 fine faces, 1121864 fine elements

Processors: 8
---------------------
Fine Matrix
---------------------
Size            36052000
NonZeros:       134788000

Coarse Matrix
---------------------
Size            605484
NonZeros:       600930

Op Comp:        1.17

Upscale Setup Time:      194.7

Coarse Solve Time:       287.991
Coarse Solve Iterations: 2688

Fine Solve Time:         57.52
Fine Solve Iterations:   22
{
  "finest-div-error": 0.32307985763419617,
  "finest-p-error": 0.43528911715481983,
  "finest-u-error": 0.64475987000378587,
  "operator-complexity": 1.1725354944728636
}

real    9m41.518s
user    74m19.232s
sys     1m35.285s
```
vs Allowing some unbalance
```
time mpirun -np 8 ./finitevolume --dim 3 -m 4 -t .10 -ma -hb
// ...  same as above 
Coarse Matrix
---------------------
Size            604716
NonZeros:       600234

Op Comp:        1.17

Upscale Setup Time:      194.979

Coarse Solve Time:       213.376
Coarse Solve Iterations: 1981

Fine Solve Time:         54.3113
Fine Solve Iterations:   22
{
  "finest-div-error": 0.34983192389957152,
  "finest-p-error": 0.40914296492415175,
  "finest-u-error": 0.62515876852662322,
  "operator-complexity": 1.1718853642241089
}

real    8m29.240s
user    64m58.563s
sys     1m49.336s
```
 In this case, it still does not beat the fine solver, but an non-insignificant improvement vs the previous partitioning.

I'm not sure if there are downsides to allowing imbalance.  The setup time seems to stay about the same.  A concern might be that it may produce empty partitions for some edge cases (where `num_procs` is close to `num_elements`). 

This may not be better in every scenario so perhaps this could be user set?

### DIFF
--- a/examples/finitevolume.cpp
+++ b/examples/finitevolume.cpp
@@ -75,6 +75,9 @@ int main(int argc, char* argv[])
     args.AddOption(&metis_agglomeration, "-ma", "--metis-agglomeration",
                    "-nm", "--no-metis-agglomeration",
                    "Use Metis as the partitioner (instead of geometric).");
+    double proc_part_ubal = 2.0;
+    args.AddOption(&proc_part_ubal, "-pub", "--part-unbalance",
+                   "Processor partition unbalance factor.");
     int spe10_scale = 5;
     args.AddOption(&spe10_scale, "-sc", "--spe10-scale",
                    "Scale of problem, 1=small, 5=full SPE10.");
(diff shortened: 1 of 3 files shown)


chakshinglee (approved):

barker29 (approved):
I am a tiny bit concerned this is premature optimization - sacrificing simplicity for a little performance. But the performance is noticeable and the complexity is small, so we can go ahead with it.

PR #14, opened 2018-03-01
Allow changing coefficient (edge weight) after coarsening [feature/coarse-mass]

barker29:
The idea here is to construct the coarse space (especially the coarse mass matrix) in such a way that you can change/rescale the edge weights after coarsening, without recoarsening. This may be useful in certain UQ problems where we want to run problems at different resolutions with different coefficients.

This is incomplete, but I think it's time to get some feedback on it.

TODO:
- [x] implement in parallel (currently only works in serial)
- [x] use edge weights on fine level in construction (currently assumes edge weights of 1)
- [x] change the interface so we can actually use the key feature, changing coefficients without re-coarsening.

Items for discussion or future development:
- figure out how this works with hybridization (not simple, will consider later)
- replace standard assembly with this? Do we need to keep both? (will reconsider later)
- this doesn't really fit with `CoarseMBuilder` interface, maybe refactor (again, let's put this off)
- I am not sure storing the components as `std::vector<mfem::DenseMatrix>` (for now this is okay, @chakshinglee has an idea of how to reduce the number of these we store)


src/GraphCoarsenBuilder.cpp
```diff
@@ -201,6 +260,239 @@ std::unique_ptr<mfem::SparseMatrix> AssembleMBuilder::GetCoarseM()
...
+
+    // ---
+    // assemble from components...
+    // ---
+    auto CoarseM = make_unique<mfem::SparseMatrix>(
+                       total_num_traces_ + ncoarse_vertexdofs_ - num_aggs,
+                       total_num_traces_ + ncoarse_vertexdofs_ - num_aggs);
+
+    // F_F block, the P_F^T M_F P_F pieces (this explicit transpose is not my favorite)
+    auto face_Agg = smoothg::Transpose(topology_.Agg_face_);
```
gelever: `GraphTopology` has `face_Agg_` ? Unless that's some how different.

src/GraphUpscale.cpp
```diff
@@ -23,22 +23,25 @@
 namespace smoothg
 {
 
+/// @todo why is there not timing here?
```
gelever, line 26: Good catch, the timing is not consistent between the constructors.  It should be moved outside of the `Init` function, or move the `chrono.Stop();` block in the coarsening constructor up before the `Init` call (lines 62-63 moved up to 61-62).

src/utilities.hpp
```diff
@@ -250,6 +250,10 @@ class ParGraph
 void GetTableRow(
     const mfem::SparseMatrix& mat, int rownum, mfem::Array<int>& J);
 
+/// if you call GetTableRow repeatedly, bad things might happen
```
gelever, line 253: Just curious, what issues did you run into with this?

chakshinglee:
@barker29 Are the components of coarse M (M_{EF}, M_{EE}, M_{FF}) kept somewhere. I think they need to be kept so that when the coefficient is updated, the coarsen M can be updated by assembling those components with certain scaling (based on the updated coefficient).

barker29:
Yes, the components are kept in the `CoefficientMBuilder` class, see `GraphCoarsenBuilder.hpp` around line 242. I think the reconstruction with new coefficient can be done now, even though it is not tested.

src/GraphCoarsenBuilder.cpp
barker29: You're right, not sure how I missed that. Fixed in last commit.

src/GraphCoarsen.cpp
```diff
@@ -298,12 +299,18 @@ void GraphCoarsen::BuildPEdges(
     NormalizeTraces(edge_traces, Agg_vertex, face_edge);
 
     std::unique_ptr<CoarseMBuilder> mbuilder_ptr;
```
chakshinglee: In order to reuse the components, the scope of `mbuilder_ptr` needs to be in `GraphCoarsen`, not just in this function? (this is probably in your TODO)

barker29: Yes, that's what I mean by the third TODO item. I'm not sure exactly how to do it, but probably pass `mbuilder` to `BuildPEdges` instead of constructing it there. Right now I'm just demonstrating that this way of constructing coarse M works the same as the previous way.

barker29:
The Travis failures are because `SparseMatrix::GetSubMatrix()` is const for the MFEM version I'm using but not the one on Travis. I'll fix this soon.

barker29:
I think this has all the basic functionality now. I'd be happy to see people's thoughts on the interface/implementation etc., but I think it's close to being ready to merge.

src/GraphCoarsenBuilder.hpp
```diff
@@ -148,10 +164,21 @@ class ElementMBuilder : public CoarseMBuilder
...
-    std::unique_ptr<mfem::SparseMatrix> GetCoarseM();
+    /// Here returns a null pointer
+    /// @todo change interface so this is optional?
+    std::unique_ptr<mfem::SparseMatrix> GetCoarseM(
+        const mfem::Vector& fineMdiag,
+        const mfem::SparseMatrix& Pedges, const mfem::SparseMatrix& face_cdof);
+
+    bool NeedsCoarseVertexDofs() { return true; }
+
+    /// @todo does DenseMatrix have a move constructor? will this do a deep copy?
```
gelever: This shouldn't deep copy unless the caller requests a copy, right?
``` c++
const std::vector<DenseMatrix>& CM_el = element_builder.GetElementMatrices(); // No copy
std::vector<DenseMatrix> CM_el_copy = element_builder.GetElementMatrices(); // Deep copy
```
Where the move constructor is never involved, `ElementMBuilder` always keeps its version of the data.

examples/mlmc.cpp
```diff
@@ -0,0 +1,351 @@
...
+    fvupscale.PrintInfo();
+    fvupscale.ShowSetupTime();
+
+    mfem::BlockVector rhs_fine(fvupscale.GetFineBlockVector());
+    rhs_fine.GetBlock(0) = 0.0;
+    rhs_fine.GetBlock(1) = rhs_u_fine;
+
+    const int num_fine_edges = vertex_edge.Width();
+    const int num_aggs = partitioning.Max() + 1; // this can be wrong if there are empty partitions
+    StupidSampler sampler(num_fine_edges, num_aggs);
```
chakshinglee: @barker29 this is a great name! Do you want to keep this name or use something like `SimpleSampler`?

src/GraphCoarsenBuilder.cpp
```diff
@@ -69,6 +93,10 @@ void AssembleMBuilder::RegisterRow(int agg_index, int row, int cdof_loc, int bub
     bubble_counter_ = bubble_counter;
 }
 
+void CoefficientMBuilder::RegisterRow(int agg_index, int row, int cdof_loc, int bubble_counter)
+{
+}
```
gelever: Instead of all of these empty implementations, the base class `CoarseMBuilder` can just have the empty implementation (instead of being pure virtual).  Then derived classes can override the implementation if they need something specific done.  Maybe keeping `Setup` as the only pure virtual function.  Should simplify the current design a bit.

src/FiniteVolumeUpscale.cpp
```diff
@@ -51,9 +51,21 @@ FiniteVolumeUpscale::FiniteVolumeUpscale(MPI_Comm comm,
     auto graph_topology = make_unique<GraphTopology>(ve_copy, edge_d_td_, global_partitioning,
                                                      &edge_boundary_att_);
 
+    std::shared_ptr<CoarseMBuilder> mbuilder_ptr;
+    std::shared_ptr<ElementMBuilder> hybrid_builder_ptr;
```
gelever, line 55: Can these be collapsed into single `std::unique_ptr`?

src/GraphCoarsenBuilder.cpp
```diff
@@ -190,17 +243,258 @@ void AssembleMBuilder::AddTraceAcross(int row, int col, double value)
...
+    Rt.RightScaling(D);
+    Mult(Rt, P, out);
+    return out;
+}
+
+void CoefficientMBuilder::BuildComponents(const mfem::Vector& fineMdiag,
+                                          const mfem::SparseMatrix& Pedges,
+                                          const mfem::SparseMatrix& face_cdof)
+{
+    // in future MFEM releases when SparseMatrix::GetSubMatrix is const-correct,
```
gelever, line 277: Maybe tag w/ `todo` or similar so it will be easier to find later

src/GraphCoarsenBuilder.cpp
```diff
@@ -190,17 +243,258 @@ void AssembleMBuilder::AddTraceAcross(int row, int col, double value)
...
+            for (int fprime = f; fprime < local_faces.Size(); ++fprime)
+            {
+                int faceprime = local_faces[fprime];
+                GetCoarseFaceDofs(face_cdof, faceprime, coarse_face_dofs_prime);
+                AddScaledSubMatrix(*CoarseM, coarse_face_dofs,
+                                   coarse_face_dofs_prime, comp_EF_EF_[counter],
+                                   agg_weight);
+                if (f != fprime)
+                {
+                    mfem::DenseMatrix EFEFtrans(comp_EF_EF_[counter], 't');
```
gelever, line 429: The explicit copy and transpose can be avoided by inserting the submatrix transposed, but I'm not sure that it is worth the hassle.

src/FiniteVolumeMLMC.cpp
```diff
@@ -0,0 +1,153 @@
...
+    setup_time_ += chrono.RealTime();
+}
+
+/// this implementation is sloppy
+void FiniteVolumeMLMC::RescaleFineCoefficient(const mfem::Vector& coeff)
+{
+    mfem::Vector temp(coeff);
+    for (int i = 0; i < coeff.Size(); ++i)
+        temp(i) = 1.0 / coeff(i);
+    GetFineMatrix().ScaleM(temp);
```
chakshinglee: Like the coarse level update, the fine level will need to keep data of the original M, and produce a new M whenever an element scaling coefficient is given. The current way will rescale the rescaled matrix. So in mlmc.cpp, if more than 2 samples are tested, the coarse and fine level pressure will get different solutions

examples/mlmc.cpp
barker29: Yes, that's more professional. Done.

src/GraphCoarsenBuilder.cpp
barker29: Thanks, done.

src/FiniteVolumeUpscale.cpp
barker29: Probably, but it would require a little reorganizaion. I need a `ElementMBuilder` pointer to construct the `HybridSolver`, but I can't build this kind of pointer if hybridization is turned off. There are several solutions here, but this seemed the simplest.

src/GraphCoarsenBuilder.cpp
barker29: Done.

src/GraphCoarsenBuilder.cpp
barker29: That's not a bad idea, but it would involve nontrivial modifications to `AddScaledSubMatrix`. At this stage I don't think it's worth the hassle.

src/FiniteVolumeMLMC.cpp
barker29: Good catch, this is now fixed.

src/GraphCoarsenBuilder.cpp
```diff
@@ -190,17 +210,255 @@ void AssembleMBuilder::AddTraceAcross(int row, int col, double value)
...
+
+mfem::DenseMatrix CoefficientMBuilder::RTDP(const mfem::DenseMatrix& R,
+                                            const mfem::Vector& D,
+                                            const mfem::DenseMatrix& P)
+{
+    mfem::DenseMatrix out(R.Width(), P.Width());
+    mfem::DenseMatrix Rt;
+    Rt.Transpose(R);
+    Rt.RightScaling(D);
+    Mult(Rt, P, out);
```
gelever: MFEM w/ LAPACK will break if these have a dimension of size zero.  We can check for this and return early.

src/GraphCoarsenBuilder.cpp
```diff
@@ -190,17 +210,255 @@ void AssembleMBuilder::AddTraceAcross(int row, int col, double value)
...
+    // EF_E block and E_E block
+    comp_E_E_.resize(num_aggs);
+    for (int agg = 0; agg < num_aggs; ++agg)
+    {
+        GetCoarseAggDofs(agg, local_coarse_dofs);
+        if (local_coarse_dofs.Size() == 0)
+        {
+            mfem::DenseMatrix empty(0, 0);
+            comp_E_E_[agg] = empty;
+            for (int af = 0; af < local_faces.Size(); ++af)
```
gelever, line 336: This loop's stop condition `local_faces.Size()` is not set for the current aggregate.  It will use whatever size was previously set.  The test cases don't cover this scenario.  
There is probably a much simpler set of parameters to make this crash, but the following  segfaults on my machine
```sh
./generalgraph -m 4  -t 0.01 --coarse-coefficient -gg -nv 5000 -np 1000 -gf -ma
```

gelever (changes requested):

src/GraphCoarsenBuilder.cpp
barker29: Thanks, good catch, I've fixed these two bugs.

gelever (approved):

barker29:
@chakshinglee do you have suggestions or do you think this PR is ready to merge?

osborn9 (approved):

chakshinglee:
@barker29 I think it is ready. 

Only one minor issue: the fine scale rescaling is given by a vector in the edge space, while in the coarse scale the rescaling vector is element-based. I think it would be nice to have consistent setting. I am not sure if this can be done, but my understanding is that element scaling is what is generated (by solving some PDE or something else), so users probably need to do this conversion step or we do it for them. This might be considered later.

chakshinglee (approved):

PR #16, opened 2018-04-18
Update coarse hybrid solver when edge weights are rescaled

chakshinglee:
This PR is a follow up to PR #14. One of the TODOs there is to build coarse M when hybridization is used as solver, and the main purpose of this PR is for that.

In particular, I made some changes to `HybridSolver` so that when "aggregate weight" (think of it as element scaling in the case of finite volume) is updated, the solver will rebuild with minimal effort. To do this, the element matrices will be stored in `Hybrid_el` (previously we only do this when SA-AMGe is called), and the hybridized system will be assembled by adding Hybrid_el scaled by aggregate weight to the global matrix.

Another change is that `AssembleMBuilder` is removed (following @barker29 's suggestion). If the global assembled M is needed, it can be assembled using the element matrices in `ElementMBuilder`, so we are not losing functionality.

There is one last thing in this PR: to make "element matrices" for the fine level as well. I dont know if this is a good idea, but I just wanted to have some consistent treatment. So for finite volume problems, we could construct element matrices, and assemble the global M using local matrices and "vertex_edge". We can use the idea of current `CoarseMBuilder`, make one more abstraction as `MBuilder`. This class has the ability to return the global assembled M when requested, and optionally it can return element matrices. So it can be used not only in `GraphCoarsen`, but in general it replaces the matrix M. Currently I dont know how to make sense of "local weight" in the sense of graph, but for our previous case in which edge weight is given, it is treated similar to how we distribute weight in parallel.


src/FiniteVolumeUpscale.hpp
```diff
@@ -57,7 +57,7 @@ class FiniteVolumeUpscale : public Upscale
     FiniteVolumeUpscale(MPI_Comm comm,
                         const mfem::SparseMatrix& vertex_edge,
                         const mfem::Vector& weight,
-                        const mfem::Array<int>& global_partitioning,
+                        const mfem::Array<int>& partitioning,
```
barker29, line 60: You changed this name here but not in `GraphUpscale`, is that intentional?

src/SpectralAMG_MGL_Coarsener.hpp
```diff
@@ -56,6 +56,7 @@ class SpectralAMG_MGL_Coarsener : public Mixed_GL_Coarsener
               HybridSolver
        @param coarse_coefficient whether to construct coarse mass matrix in a way
               so the coefficients (edge weights) can be rescaled after coarsening
+              (this documentation needs to be updated)
```
barker29: What better time to update it than now?

src/GraphCoarsen.cpp
```diff
@@ -508,7 +514,14 @@ void GraphCoarsen::BuildPEdges(
...
 
-    CoarseM_ = mbuilder.GetCoarseM(M_v, Pedges, face_cdof);
+    auto coef_mbuilder_ptr = dynamic_cast<CoefficientMBuilder*>(&mbuilder);
+    if (coef_mbuilder_ptr)
+    {
+        coef_mbuilder_ptr->BuildComponents(M_v, Pedges, face_cdof);
+    }
+
+    // TODO: only compute CoarseM_ when solver is not hybridization
+    CoarseM_ = mbuilder.GetAssembledM();
```
barker29: I'm trying to track this down, and I don't think this `CoarseM_` is actually used at all (no matter the solver). Instead, I think we always use `mbuilder->GetAssembledM()` via `MixedMatrix::UpdateM()`. Can you try to verify this? I think this also means that most of our `MixedMatrix` constructors are unused (only the one that takes an `MBuilder` is used). Cleaning up some of this is probably out of scope for this PR but should be done at some point.

src/GraphCoarsenBuilder.hpp
```diff
@@ -28,6 +28,33 @@
 namespace smoothg
 {
 
+class MBuilder
```
barker29: Can we get some documentation, even just a single line, for this class?

src/HybridSolver.cpp
```diff
@@ -85,20 +85,18 @@ HybridSolver::HybridSolver(MPI_Comm comm,
     mfem::SparseMatrix tmp = SparseIdentity(nvertices);
     Agg_vertexdof_.Swap(tmp);
 
-    Agg_edgedof_.MakeRef(D_);
     const mfem::SparseMatrix edge_edgedof;
 
-    std::vector<mfem::Vector> M_el;
-    BuildFineLevelLocalMassMatrix(D_, mgL.getWeight(), M_el);
+    auto& mbuilder = static_cast<const FineMBuilder&>(mgL.GetMBuilder());
```
barker29: Can this `static_cast` fail, ie if `GetMBuilder` returns something that's not actually a `FineMBuilder`? It might be better to dynamic_cast and check the return. @gelever can you give an opinion?

src/MixedMatrix.hpp
```diff
@@ -104,14 +105,46 @@ class MixedMatrix
...
+        assert(M_ || mbuilder_);
+        if (!M_)
+        {
+            M_ = mbuilder_->GetAssembledM();
+        }
         return *M_;
     }
 
+    /**
+       @brief Get a reference to the mass matrix M.
```
barker29: Correct this documentation.

src/MixedMatrix.hpp
```diff
@@ -304,7 +350,7 @@ class MixedMatrix
 
     void GenerateRowStarts();
 
-    std::unique_ptr<mfem::SparseMatrix> M_;
+    mutable std::unique_ptr<mfem::SparseMatrix> M_;
```
barker29: I'm not sure this should actually be mutable - if we are changing `M_` we are really changing the object, not just some implementation detail. Maybe remove `const` from this object in situations where we rescale the mass matrix?

src/MatrixUtilities.hpp
```diff
@@ -397,6 +397,8 @@ double InnerProduct(const mfem::Vector& u, const mfem::Vector& v);
 std::unique_ptr<mfem::HypreParMatrix> BuildEntityToTrueEntity(
     const mfem::HypreParMatrix& entity_trueentity_entity);
 
+void BooleanMult(const mfem::SparseMatrix& mat, const mfem::Array<int>& vec,
```
barker29: Some brief documentation please.

src/HybridSolver.hpp
```diff
@@ -226,6 +236,8 @@ class HybridSolver : public MixedLaplacianSolver
     // Construct spectral AMGe preconditioner
     void BuildSpectralAMGePreconditioner();
 
+    void BuildParallelSystemAndSolver();
```
barker29: A little documentation please.

barker29 (changes requested):
This is overall good work and a significant improvement to the code - using hybridization with rescaled coefficients will be important. That said, I have a few relatively minor issues I'd like to see addressed before it's merged.

src/GraphCoarsen.cpp
barker29: Shorter version: I think you can remove lines 523-524 now, and later on we should figure out why.

src/HybridSolver.cpp
gelever: I'm not sure what will happen.  I think it's [UB](http://eel.is/c++draft/expr.static.cast#2) if it fails (?).  Dynamic cast will definitely catch an incorrect cast.

src/FiniteVolumeUpscale.hpp
chakshinglee: Yes, `partitioning` here is for the local graphs of individual processors, while `global_partitioning` in `GraphUpscale` is for the global graph before distribution. 

This is because of the setup of `ParGraph`, which needs to be changed (now the global graph is partitioned and distributed based on global partition, it should be better to distribute first and then partition local graphs). But I plan to change it in this PR.

Edited: In the last sentence, I wanted to say I dont plan to change it in this PR

src/SpectralAMG_MGL_Coarsener.hpp
chakshinglee: Right, I will update it now (I actually didn't know how to characterize this parameter, but I think I know now).

src/GraphCoarsen.cpp
chakshinglee: Good catch! Yes, I think this is because the role of the M matrix is replaced by `mbuilder` now, and like you said the M matrix will be constructed using `mbuilder->GetAssembledM()` (this is also called in `MixedMatrix::getWeight()`).

src/MixedMatrix.hpp
chakshinglee: This is a good point. I think I make `M_` mutable mainly because of `MixedMatrix::getWeight() const`. This function should really just return a reference to `*M_`. Currently `M_` is changed in  getWeight() only if it is `nullptr` (similar to `get_pM()` and `get_pD()`). In the case when `M_` is rescaled, it is updated in `UpdateM()`, so `getWeight()` would not change `M_` because it is not `nullptr`. So "technically" the mass matrix is not changed in either case. But I agree this should be better designed.

src/HybridSolver.cpp
chakshinglee: Thanks for the suggestion, yes, I agree dynamic_cast is a better choice.

chakshinglee:
@barker29 thanks for checking out this PR, your suggestions definitely make this a better PR. I have made some commits based your suggestions, please check the updated version when you have time.

src/SpectralAMG_MGL_Coarsener.hpp
```diff
@@ -54,8 +54,8 @@ class SpectralAMG_MGL_Coarsener : public Mixed_GL_Coarsener
               (guarantees approximation property in edge energy norm)
        @param is_hybridization_used whether to prepare the coarse space to use the
               HybridSolver
-       @param coarse_coefficient whether to construct coarse mass matrix in a way
-              so the coefficients (edge weights) can be rescaled after coarsening
+       @param coarse_components whether to store different components of coarse M
+              based on trace extensions and bubbles (for construction of coarse M)
```
chakshinglee, line 58: @barker29 This parameter is renamed because now both `CoefficientMBuilder` and `ElementMBuilder` have the ability to compute the rescaled coarse M (without going back to the fine grid). The difference between the two is, the former stores various components of coarse M (M_{FF}, M_{FE}, etc.), while the latter stores only the element matrices. I believe now `ElementMBuilder` is capable of doing what `CoefficientMBuilder` can do, up to the issue (about harmonic average and normal average) that we talked about some time ago.

barker29: Great, do you recommend removing `CoefficientMBuilder` in the future?

barker29 (approved):

examples/mlmc.cpp
```diff
@@ -209,14 +216,32 @@ int main(int argc, char* argv[])
...
+                local_weight_i.SetSize(M_el_i.Height());
+                for (int j = 0; j < local_weight_i.Size(); j++)
+                {
+                    local_weight_i[j] = 1.0 / M_el_i(j, j);
+                }
+            }
+        }
     }
 
     for (int i = 0; i < weight.Size(); ++i)
```
gelever: This loop should also probably be moved into the `if (elem_mass == false)` block.

src/FiniteVolumeMLMC.cpp
```diff
@@ -73,80 +124,109 @@ FiniteVolumeMLMC::FiniteVolumeMLMC(MPI_Comm comm,
...
-        temp(i) = coeff(i) * weight_(i);
-    GetFineMatrix().SetMFromWeightVector(temp);
+    if (!hybridization_)
+    {
+        GetFineMatrix().UpdateM(coeff);
+        ForceMakeFineSolver();
+    }
+    else
+    {
+        ((HybridSolver&) (*fine_solver_)).UpdateAggScaling(coeff);
```
gelever: I think it is generally recommended to avoid C style casts.

src/HybridSolver.cpp
```diff
@@ -666,15 +577,16 @@ void HybridSolver::Mult(const mfem::BlockVector& Rhs, mfem::BlockVector& Sol) co
...
+    timing_ = chrono.RealTime();
 
     if (myid_ == 0 && print_level_ > 0)
     {
         std::cout << "  Timing: PCG done in "
                   << timing_ << "s. \n";
     }
 
-    num_iterations_ += cg_->GetNumIterations();
+    // TODO: decide to use = or += here and in timing_ update (MinresBlockSolver uses +=)
```
gelever, line 588: Perhaps total iterations should be tracked separately. Or maybe not at all and let the user take care of it.

gelever (approved):
Looks great ! :+1:

examples/mlmc.cpp
chakshinglee: Good catch!

src/FiniteVolumeMLMC.cpp
chakshinglee: Right, I will change it to dynamic_cast, thanks!

src/HybridSolver.cpp
chakshinglee: I think it's more like the user's job to keep track of the total iteration/time if the solver is called in a loop.

src/HybridSolver.cpp
```diff
@@ -855,7 +855,7 @@ void HybridSolver::BuildParallelSystemAndSolver()
         else
         {
             prec_ = make_unique<mfem::HypreBoomerAMG>(*pHybridSystem_);
-            ((mfem::HypreBoomerAMG&)*prec_).SetPrintLevel(0);
+            static_cast<mfem::HypreBoomerAMG&>(*prec_).SetPrintLevel(0);
```
chakshinglee: @gelever @barker29 Is `static_cast` more preferred here?

gelever: [CppCoreGuidelines](https://github.com/isocpp/CppCoreGuidelines/blob/d49c720e98a914de4a21e6b8618dd4102c0a0139/CppCoreGuidelines.md#c146-use-dynamic_cast-where-class-hierarchy-navigation-is-unavoidable) recommends dynamic cast even if you know the type it will be.

barker29: I took the liberty of fixing this, removing the cast entirely. If you don't like it I'll revert.

chakshinglee: @gelever That's interesting. Based on the guidelines, seems there are really really rare cases one will use static_cast. 

@barker29 That looks fine to me, thanks!

PR #17, opened 2018-04-18
Fix zero size mult, add all tests to travis

gelever:
When checking out #16, I noticed there was a failing test `isolate-coarsen`.  ~~This is actually broken for me on master too.~~  Only broken w/ #16, not master.

This PR allows the test suite to have a variable number of processors and tolerance. Travis CI has 2 cores. When overcommitting w/ 4 cores the tests takes forever to complete.  Setting it to 2 cores lets us run all tests and it will complete in a reasonable time.  But the tolerance is lower to account for the different number of cores.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,24 +11,12 @@ matrix:
       compiler: gcc
       addons:
         apt:
-          # sources:
-            # - ubuntu-toolchain-r-test
           packages:
-            # GCC 4.9
-            # - g++-4.9
-            # MPICH
-            #- mpich
-            #- libmpich-dev
-            # Other
             - liblapack-dev
             - libblas-dev
             - doxygen
             - valgrind
             - gfortran
-            # OpenMPI
-            #- gfortran
-            #- openmpi-bin
-            #- libopenmpi-dev
       env: DEBUG=YES
            MPI=YES
       cache:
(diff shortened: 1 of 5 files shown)


chakshinglee (approved):
@gelever this looks nice to me, thanks!

.travis.yml
```diff
@@ -167,15 +155,13 @@ script:
      CPPFLAGS="";
      SKIP_TEST_DIRS="";
 
-
    # Configure the library
    - mkdir build;
    - cd build;
-   - CC=gcc CXX=g++ cmake -DMFEM_DIR="/home/travis/build/LLNL/mfem-install" -DHYPRE_DIR="/home/travis/build/LLNL/hypre-install" -DSuiteSparse_DIR="/home/travis/build/LLNL/SuiteSparse-install" -DMETIS_DIR="/home/travis/build/LLNL/metis-install" -DUSE_ARPACK=OFF -DCMAKE_BUILD_TYPE=DEBUG -DSPE10_PERM="/home/travis/build/LLNL/spe10-install/spe_perm.dat" ..;
+   - CC=mpicc CXX=mpic++ cmake -DMFEM_DIR="/home/travis/build/LLNL/mfem-install" -DHYPRE_DIR="/home/travis/build/LLNL/hypre-install" -DSuiteSparse_DIR="/home/travis/build/LLNL/SuiteSparse-install" -DMETIS_DIR="/home/travis/build/LLNL/metis-install" -DUSE_ARPACK=OFF -DCMAKE_BUILD_TYPE=Debug -DSPE10_PERM="/home/travis/build/LLNL/spe10-install/spe_perm.dat" -DSMOOTHG_TEST_PROCS=2 -DSMOOTHG_TEST_TOL="1e-2" ..;
```
barker29: A very nitpicky question: can this very long line be split up without ruining the YAML?

examples/graphupscale.cpp
```diff
@@ -165,9 +165,11 @@ int main(int argc, char* argv[])
 
 #if SMOOTHG_USE_SAAMGE
         SAAMGeParam saamge_param;
+        const bool coarse_coefficient = false;
         const auto hbsa_upscale = GraphUpscale(comm, vertex_edge, coarse_factor,
                                                spect_tol, max_evects, dual_target,
                                                scaled_dual, energy_dual, use_hybridization,
+                                               coarse_coefficient,
```
barker29, line 172: I added this argument so the code will compile with SAAMGe enabled. It doesn't really belong in this PR, let me know if it's an issue and I'll remove it and issue a separate PR.

barker29 (approved):

examples/graphupscale.cpp
gelever: Its fine here :+1:

.travis.yml
gelever: Done for this line.  Idk if the `|` character will work for the other long lines.

PR #18, opened 2018-04-27
Fix a null fine M

chakshinglee:
When `--element-mass` option is used in mlmc.cpp, `M_` in the fine level `MixedMatrix` will not be constructed (it is nullptr). So the assert in `MixedMatrix::GetM()` will fail.


barker29 (approved):
Looks good to me!

chakshinglee:
@barker29 sorry that I added a test for the rescaling methods after your review. Can you please take a look at that also, thanks! 

The test is to verify that M obtained from rescaling is same as direct assembling using coefficient.

barker29:
Still looks good to me. I like the rescaling test.

gelever (approved):
Everything builds and passes tests on my end :+1:

PR #19, opened 2018-05-04
Improve/elem coarse m

chakshinglee:
While I was trying to use the rescaling feature in the two-phase flow code, where the M matrix will need to be updated during time stepping, I found that the coarse solution is not the same for serial and parallel run. So I trace back and realize the problem is when we compute the element matrices of coarse M, we used simple splitting (divided by 2) for the submatrices of M on faces (which is possibly the best one can do if only assembled fine M is given). But in the case when element matrices of fine M are available (through `FineMBuilder`), we can make use of it in the coarsening so that rescaling works for more general coefficients. This is the main purpose of this PR.

The test `rescaling.cpp` is also modified. It was testing rescaling fine/coarse M which was originally coming from finite volume problem with constant coefficient on uniform grid (note that in the case the divided by 2 strategy is doing the right job). The test now is modified to test rescaling fine/coarse M which is coming from FV problems with non-constant coefficient.

### DIFF
--- a/src/GraphCoarsen.cpp
+++ b/src/GraphCoarsen.cpp
@@ -38,24 +38,16 @@ namespace smoothg
 
   Maybe use the example in parelag-notes.
 */
-GraphCoarsen::GraphCoarsen(const mfem::SparseMatrix& M_proc,
-                           const mfem::SparseMatrix& D_proc,
-                           const GraphTopology& graph_topology)
-    : GraphCoarsen(M_proc, D_proc, nullptr, graph_topology)
-{
-}
-
-GraphCoarsen::GraphCoarsen(const mfem::SparseMatrix& M_proc,
-                           const mfem::SparseMatrix& D_proc,
-                           const mfem::SparseMatrix* W_proc,
-                           const GraphTopology& graph_topology)
+GraphCoarsen::GraphCoarsen(const MixedMatrix& mgL, const GraphTopology& graph_topology)
     :
-    M_proc_(M_proc),
-    D_proc_(D_proc),
-    W_proc_(W_proc),
+    M_proc_(mgL.GetM()),
+    D_proc_(mgL.GetD()),
+    W_proc_(mgL.GetW()),
+    fine_mbuilder_(dynamic_cast<const FineMBuilder*>(&(mgL.GetMBuilder()))),
     graph_topology_(graph_topology),
-    colMapper_(M_proc.Size())
+    colMapper_(M_proc_.Size())
 {
+    assert(fine_mbuilder_);
     colMapper_ = -1;
 }
 
@@ -256,12 +248,38 @@ double GraphCoarsen::DTTraceProduct(const mfem::SparseMatrix& DtransferT,
     return smoothg::InnerProduct(ref_vec3, trace);
 }
 
-void GraphCoarsen::BuildPEdges(
-    std::vector<mfem::DenseMatrix>& edge_traces,
-    std::vector<mfem::DenseMatrix>& vertex_target,
-    mfem::SparseMatrix& face_cdof,
-    mfem::SparseMatrix& Pedges,
-    CoarseMBuilder& mbuilder)
+void GraphCoarsen::BuildAggregateFaceM(const mfem::Array<int>& edge_dofs,
+                                       const mfem::Array<int>& partition,
+                                       const mfem::SparseMatrix& edge_vert,
+                                       const int agg,
+                                       mfem::Vector& Mloc)
+{
+    Mloc.SetSize(edge_dofs.Size());
+    mfem::Array<int> verts;
+    mfem::Array<int> local_edge_dofs;
+    for (int i = 0; i < edge_dofs.Size(); i++)
+    {
+        int edge_dof = edge_dofs[i];
+        GetTableRow(edge_vert, edge_dof, verts);
+        int vert = (partition[verts[0]] == agg) ? verts[0] : verts[1];
+        const mfem::Vector& M_el_i = fine_mbuilder_->GetElementMatrices()[vert];
+        GetTableRow(fine_mbuilder_->GetAggEdgeDofTable(), vert, local_edge_dofs);
+        for (int j = 0; j < local_edge_dofs.Size(); j++)
+        {
+            if (local_edge_dofs[j] == edge_dof)
+            {
+                Mloc(i) = M_el_i(j);
+                break;
+            }
+        }
+    }
+}
+
+void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
+                               std::vector<mfem::DenseMatrix>& vertex_target,
+                               mfem::SparseMatrix& face_cdof,
+                               mfem::SparseMatrix& Pedges,
+                               CoarseMBuilder& coarse_mbuilder)
 {
     // put trace_extensions and bubble_functions in Pedges
     // the coarse dof numbering is as follows: first loop over each face, count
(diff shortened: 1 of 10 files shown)


gelever:
Am I understanding the order of events correctly?:

- There exists some initial `M` from which `P_edge`, `P_vertex`, and initial `M_c` are computed.
- `M` is rescaled at a later timestep, and then element `M_c` will also need to be modified (without changing either `P`)? But only from the faces edges?

chakshinglee:
@gelever I think everything you said is right, except for the last question.

As for as I understand, the main goal of this series of PRs is to make some tool do rescaling in the following sense: given a reference fine model, build a reference coarse model using the usual way we have been doing (this is your first point). Now when the fine model is changed (so far we only consider the change in the PDE coefficient, which affects `M` only), we want to build a pretty good coarse model for the new fine model without repeating the whole expensive coarsening procedure. One way to do that is to keep the `P`s (your second point). So ideally we want to do RAP using `P_edge` the new fine `M` to build the new coarse `M`. But if the change in the coefficient is on aggregate level (the change is piecewise constant on aggregates), we can forget about the fine M, just save the coarse level element (which is fine level aggregate) M, and multiply these element matrices by the change, and assemble them into global M. 

So actually the whole `M_c` is modified, not just the face edges. But they are trickier because they are shared by two aggregates, so we will need to separate the fine M into element matrices in order to get the correct splitting (because the coefficient is defined on elements, not faces).

src/GraphCoarsen.cpp
```diff
@@ -461,23 +483,26 @@ void GraphCoarsen::BuildPEdges(
...
     {
         elem_mbuilder_ptr->SetAggToEdgeDofsTableReference(*Agg_cdof_edge_);
     }
 
     mfem::SparseMatrix face_Agg(smoothg::Transpose(Agg_face));
-    // next line assume M_proc_ is diagonal
-    mfem::Vector M_v(M_proc_.GetData(), M_proc_.Width()), Mloc_v;
-    mbuilder.ResetEdgeCdofMarkers(total_num_traces + bubble_counter);
 
+    auto edge_vert = smoothg::Transpose(D_proc_); // TODO: use vertex_edge
```
barker29: I get nervous about TODO comments in PRs. Either fix it or document exactly what's wrong?

src/GraphCoarsenBuilder.cpp
```diff
@@ -110,77 +104,46 @@ void ElementMBuilder::SetBubbleBubbleBlock(int l, int j, double value)
...
-    GetTableRow(face_Agg, face_num, Aggs);
-    Agg0_ = Aggs[0];
-    GetTableRow(Agg_cdof_edge, Agg0_, local_Agg_edge_cdof);
-    for (int k = 0; k < local_Agg_edge_cdof.Size(); k++)
-    {
-        edge_cdof_marker_[local_Agg_edge_cdof[k]] = k;
-    }
-    if (Aggs.Size() == 2)
+    GetTableRow(face_Agg, face_num, Aggs_);
+    for (int a = 0; a < Aggs_.Size(); a++)
```
barker29, line 116: These are some nice ways to simplify the code.

src/GraphCoarsenBuilder.cpp
```diff
@@ -110,77 +104,46 @@ void ElementMBuilder::SetBubbleBubbleBlock(int l, int j, double value)
...
-    {
-        mfem::DenseMatrix& CM_el_loc2(CM_el_[Agg1_]);
-        CM_el_loc1(id0_in_Agg0, id1_in_Agg0) += value / 2.;
-        int id0_in_Agg1 = edge_cdof_marker2_[row];
-        int id1_in_Agg1 = edge_cdof_marker2_[col];
-        CM_el_loc2(id0_in_Agg1, id1_in_Agg1) += value / 2.;
-    }
+    int id0_in_agg = edge_cdof_markers_[agg][row];
+    int id1_in_agg = edge_cdof_markers_[agg][col];
+    CM_el_loc(id0_in_agg, id1_in_agg) += value;
```
barker29, line 132: This all looks much simpler and cleaner, let me make sure I understand: now in this code you no longer divide by 2 for these local matrices at this stage, but when you actually assemble the matrix (around line 146) the `agg_weight` includes the appropriate scaling, so the final result is the same as before?

src/GraphCoarsenBuilder.hpp
```diff
@@ -42,10 +42,13 @@ class MBuilder
...
        @brief Build the assembled M for the local processor
      */
-    virtual std::unique_ptr<mfem::SparseMatrix> BuildAssembledM() const = 0;
+    std::unique_ptr<mfem::SparseMatrix> BuildAssembledM() const;
 
     /**
-       @brief Set weights on aggregates for assembly of mass matrix.
+       @brief Assemble the rescaled M for the local processor
+
+       Build the assembled M with "element" contribution inversely scaled
```
barker29: This makes sense to me now, but maybe for others (or for me in six months) describe briefly what "element contribution" means.

Or maybe just remove these words to avoid confusion? The interface is actually just aggs; for each agg you supply an (inverse) weight, that may be all the user needs to know.

barker29 (approved):
This looks very nice in general.

src/GraphCoarsen.cpp
chakshinglee: Right now this is fine. I was thinking this will be wrong for multilevel (although in multilevel many things in this code will be wrong anyway). I will just erase this TODO.

src/GraphCoarsenBuilder.hpp
chakshinglee: Since `MBuilder` is a general class for any level, this function works for fine level as well. For each level, this function actually takes the element (inverse) weight in that level. In particular, an element in the coarse level is an aggregate in the fine level. 

In any case, this whole thing only makes sense to me in finite volume, so I find it hard to document it in graph language. I will take your advice to remove those words.

src/GraphCoarsen.cpp
```diff
@@ -38,24 +38,16 @@ namespace smoothg
...
-    M_proc_(M_proc),
-    D_proc_(D_proc),
-    W_proc_(W_proc),
+    M_proc_(mgL.GetM()),
+    D_proc_(mgL.GetD()),
+    W_proc_(mgL.GetW()),
+    fine_mbuilder_(dynamic_cast<const FineMBuilder*>(&(mgL.GetMBuilder()))),
     graph_topology_(graph_topology),
-    colMapper_(M_proc.Size())
+    colMapper_(M_proc_.Size())
```
gelever: This should probably be set from `D_proc_.Width()` for consistency.  Is not an issue now but in multilevel `M_proc_` may not be assembled.  Or we can wait till multilevel and fix everything then.

testcode/lineargraph.cpp
```diff
@@ -266,9 +266,14 @@ int main(int argc, char* argv[])
     mfem::SparseMatrix Pu;
     mfem::SparseMatrix Pp;
     mfem::SparseMatrix face_dof; // not used in this example
-    std::vector<mfem::DenseMatrix> CM_el_;
 
-    GraphCoarsen graph_coarsen(graph.GetM(), graph.GetD(), graph_topology);
+    mfem::Vector weight(graph.GetM().Size());
+    for (int i = 0; i < weight.Size(); i++)
+    {
+        weight(i) = graph.GetM()(i, i);
```
gelever: MFEM has `graph.GetM().GetDiag(weight)` which is a bit simpler.

gelever (approved):
Looks good to me.  Builds and passes tests on my machine.

src/GraphCoarsen.cpp
chakshinglee: I am not sure at this point how should the aggregate submatrix should be obtained in the multilevel setting (which is what `colMapper_` is for), but your suggested change is harmless and simple, so I think it makes sense to do it.

testcode/lineargraph.cpp
chakshinglee: Actually, I think I wanted to do `weight(i) = 1.0 / graph.GetM()(i, i)` but then I forgot to do so (I will fix that), although it makes no difference for this example in terms of results as M_ii is 1. The inversion is just to honour the notation used in `MixedMatrix`, so someone reading the code can find this consistent, and if someday the weight in this example is no longer 1. If we dont need to do the inversion, certainly `GetDiag` should be called.

gelever: Ah ok, that makes sense why this is in a loop :+1:

PR #21, opened 2018-05-10
hierarchical PDE sampling [feature/hierarchical-sampling]

barker29:
This PR is intended to implement the PDE sampler from Osborn, Vassilevski, Villa, SIAM J. Sci. Comput., 39(5), S543–S562, adapted for the finite volume context.

The new functionality is implemented in `examples/mlmc.cpp`, which can now produce coarse and fine versions of the same lognormal permeability field.


barker29:
Don't worry about reviewing this until PR #19 is merged and reconciled with this.

examples/finitevolume.cpp
```diff
@@ -180,16 +172,16 @@ int main(int argc, char* argv[])
                               : pmesh->ElementToFaceTable();
     mfem::SparseMatrix vertex_edge = TableToMatrix(vertex_edge_table);
 
-    // Construct agglomerated topology based on METIS or Cartesion aggloemration
+    // Construct agglomerated topology based on METIS or Cartesian aggloemration
```
gelever: `aggloemration` is also misspelled, probably by me.

examples/stest.py
```diff
@@ -361,6 +361,23 @@ def make_tests():
...
+    tests["mlmc-sanity"] = \
+        [["./mlmc",
+          "--perm", spe10_perm_file],
+         {"finest-p-error": 0.10754186878360708}]
+
+    tests["mlmc-pde-sampler"] = \
+        [["./mlmc",
+          "--sampler-type", "pde",
+          "--kappa", "0.01"],
+         {"finest-p-error": 0.19865665579298755}]
```
chakshinglee: These tests (`mlmc-pde-sampler` and `par-mlmc`) are failing in my tux box. In `mlmc.cpp`, multiple samples are generated, which error will be checked? In fact, none of the samples match this error.

examples/sampler.cpp
```diff
@@ -0,0 +1,405 @@
...
+    {
+        double count = static_cast<double>(sample) + 1.0;
+        pdesampler.NewSample();
+
+        auto sol_coarse = pdesampler.GetCoarseCoefficientForVisualization();
+        auto sol_upscaled = fvupscale->Interpolate(sol_coarse);
+        for (int i = 0; i < sol_upscaled.Size(); ++i)
+            sol_upscaled(i) = std::log(sol_upscaled(i));
+        fvupscale->Orthogonalize(sol_upscaled); // can we orthogonalize on coarse grid?
+        int coarse_iterations = fvupscale->GetCoarseSolveIters();
```
chakshinglee: https://github.com/LLNL/smoothG/blob/ff065cbdbc512c92b44162ad6daed04fcbf5da81/src/MinresBlockSolver.cpp#L218

https://github.com/LLNL/smoothG/blob/ff065cbdbc512c92b44162ad6daed04fcbf5da81/src/MinresBlockSolver.cpp#L236

We haven't changed the setting in `MinresBlockSovler`, so the number of iteration and solve time will accumulate every time when the solver is called to solve a problem.

examples/sampler.cpp
```diff
@@ -0,0 +1,405 @@
...
+    for (int sample = 0; sample < num_samples; ++sample)
+    {
+        double count = static_cast<double>(sample) + 1.0;
+        pdesampler.NewSample();
+
+        auto sol_coarse = pdesampler.GetCoarseCoefficientForVisualization();
+        auto sol_upscaled = fvupscale->Interpolate(sol_coarse);
+        for (int i = 0; i < sol_upscaled.Size(); ++i)
+            sol_upscaled(i) = std::log(sol_upscaled(i));
+        fvupscale->Orthogonalize(sol_upscaled); // can we orthogonalize on coarse grid?
```
chakshinglee: I think you can do so once you have the coarse representation of constant 1 (which you have made it available in this PR). 

In fine level this is what we do:
vec -= (1^T * vec) 1

Let P be the interpolation matrix. If we have a coarse solution vec_c, and a coarse representation of fine constant 1 (denoted by 1_c = P 1), then of course we can interpolate all the coarse quantities to fine level and use the fine level orthogonalization (and this is what the above few lines is doing):
P vec_c -= ( (P 1_c)^T * (P vec_c) ) P 1_c

The above formula can be simplified because our P satisfies P^T P = I. So 
( (P 1_c)^T * (P vec_c) ) P 1_c = ( (1_c)^T P^T * P vec_c) P 1_c = P ( (1_c)^T * vec_c) 1_c

So the coarse level orthogonalization is 
vec_c -= ( (1_c)^T *  vec_c) 1_c

You can interpolate to fine level afterwards if you want.

examples/stest.py
gelever: Same here

gelever:
`src/Sampler.hpp` needs `#include <random>` on my machine to build.  Otherwise, there are compile errors about not finding the random generator stuff.

examples/stest.py
barker29: This test was intended to only check the last sample's error (that's where the python test scripts will look), which is not a very good test but at least tells you what I was intending.

Can you guys run this code snippet and tell me the result?
```C++
#include "../src/smoothG.hpp"

int main(int argc, char * argv[])
{
    smoothg::NormalDistribution nd0(0.0, 1.0, 0);
    std::cout << nd0.Sample() << " " << nd0.Sample() << " " << nd0.Sample() << std::endl;
    smoothg::NormalDistribution nd1(0.0, 1.0, 1);
    std::cout << nd1.Sample() << " " << nd1.Sample() << " " << nd1.Sample() << std::endl;
    return 0;
}
```
On my machine the result is
```
0.0708592 0.302805 1.12279
1.58275 -1.40287 -0.549746
```
Did we have a similar issue with the random graph generation? How did we make sure that was consistent across platforms?

gelever: output on my machine:
```
./examples/testss 
1.12279 0.302805 0.0708592
-0.549746 -1.40287 1.58275
```
It might be changing how `Sample()` is executed, [left to right?](https://en.cppreference.com/w/cpp/language/eval_order)

Although this shouldn't be an issue is this PR, it just calls `Sample()` when setting a vector.

gelever: Another test:
``` c++
#include "../src/smoothG.hpp"

int main(int argc, char * argv[])
{
    smoothg::NormalDistribution nd0(0.0, 1.0, 0);
    double s1 = nd0.Sample();
    double s2 = nd0.Sample();
    double s3 = nd0.Sample();
    std::cout << s1 << " " << s2 << " " << s3 << std::endl;

    smoothg::NormalDistribution nd1(0.0, 1.0, 1);
    double s4 = nd1.Sample();
    double s5 = nd1.Sample();
    double s6 = nd1.Sample();
    std::cout << s4 << " " << s5 << " " << s6 << std::endl;

    return 0;
}
```
outputs:
```
1.12279 0.302805 0.0708592
-0.549746 -1.40287 1.58275
```

gelever: When i run `mpirun -np 2 ./mlmc --sampler-type pde --kappa 0.01 --perm fake_perm.txt`, I get the same results as the test.  I'm guessing there is an issue reading the file and different permeabilities are being used.

If the user specified file is not found, that should actually throw/error (vs silently using `BlankPermeability`)

barker29: Thanks @gelever , you're right, and I don't think it has anything to do with the random numbers. Hopefully fixed.

chakshinglee: @barker29 those tests now all pass on my tux box.

src/Sampler.cpp
```diff
@@ -0,0 +1,233 @@
...
+    normal_distribution_(0.0, 1.0, seed),
+    fine_vector_size_(vertex_edge.Height()),
+    num_coarse_aggs_(partitioning.Max() + 1),
+    cell_volume_(cell_volume),
+    current_state_(NO_SAMPLE)
+{
+    mfem::SparseMatrix W_block = SparseIdentity(vertex_edge.Height());
+    W_block *= cell_volume_ * kappa * kappa;
+    mfem::Vector one_weight(vertex_edge.Width());
+    one_weight = 1.0;
```
chakshinglee: If "kappa^2 u - \triangle u = g" is the pde considering here, the weight will need to take geometry (like face area, characteristic length) into account?

chakshinglee:
In `mlmc.cpp`, using the two solvers give different solutions (actually only edge solution is different). This is probably due to some bugs introduced by me in previous PRs. I haven't figured out what is the problem yet.

src/Sampler.cpp
barker29: Yes. The right thing to do is to use `FiniteVolumeMassIntegrator`, but that requires a BilinearForm and a Mesh, and that seems like a lot of machinery to bring into this class, and I am lazy. But probably I should just do it, since this class only makes sense in the finite volume context anyway.

barker29:
@chakshinglee do the two solvers give different results even for the first sample? Or just on later samples?

chakshinglee:
@barker29 the original tests in `stest.py` are fine (they all passed). But when I use the pde sampler in `mlmc.cpp`, the solutions become different. 

And I think I've figured out the problem. Actually, the two solvers produce the same solution (up the solver tolerance), but the fine M matrix is not updated when `HybridSolver` is called, so the "error" shown appears to be different because we measure the error in the M norm. This also explains why the test with `SimpleSampler` was fine because the rescaled coefficient is a multiple of the original coefficient only, so relative errors will cancel the multiple anyway.

https://github.com/LLNL/smoothG/blob/c183e43744c7b43ea1964298deb346f6d6070395/src/FiniteVolumeMLMC.cpp#L123-L136
Assembled fine M needs to be updated regardless which solver is called. So this line should be moved outside of the if statement. 
EDIT: I meant line 127.

src/Sampler.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+    MFEM_ASSERT(current_state_ == FINE_SAMPLE ||
+                current_state_ == COARSE_SAMPLE,
+                "PDESampler object in wrong state (call NewSample() first)!");
+
+    if (current_state_ == FINE_SAMPLE)
+        fvupscale_->Restrict(rhs_fine_, rhs_coarse_);
+    mfem::Vector coarse_sol = fvupscale_->GetCoarseVector();
+    fvupscale_->SolveCoarse(rhs_coarse_, coarse_sol);
+    coarse_sol *= -1.0; // ?
```
chakshinglee: Compare with `GetFineCoefficient`, the solution here is not orthogonalized to constants. I dont know if that is necessary, I just notice the difference between the two (and we have the tool to do the average).

src/GraphCoarsen.cpp
```diff
@@ -94,6 +98,25 @@ void GraphCoarsen::BuildPVertices(
             {
                 Pvertices_j[ptr] = coarse_vertex_dof_counter + k;
                 Pvertices_data[ptr++] = target_i(j, k);
+                if (k == 0)
+                {
+                    const double cval = coarse_constant_rep_(coarse_vertex_dof_counter);
+                    if (cval == 0.0)
+                    {
+                        // @todo this should probably depend on volume size etc?
```
chakshinglee: It may not depend on the volume of elements in fine level (unless the basis for the fine level is not constant 1 on an element), but it will depend on the volume of aggregate, which I guess is why you take `1.0 / target_i(j, k)` here.

An alternative way to compute the representation of constant vector in coarse level is to make use the fact that `Pvertices` is orthogonal, so

constant_one_coarse = Pvertices * constant_one_fine 

Then the volume factor will be taken care of automatically, and it should be extensible to multilevel as long as we construct `Pvertices` properly.

chakshinglee (approved):
@barker29 This PR looks good to me, in particular, I think the PDE sampler is very cool. There is only one thing I want to change, which is due to my mistake: 
- [ ] Enforce update of assembled fine M in both solver options so that printed errors are consistent.

There are other suggestions in the individual comments, some of them I am not sure if it is necessary for illustration purpose (e.g. taking average zero in the coarse coefficient generation, it makes sense to have similar settings in both levels though).

examples/mlmc.cpp
```diff
@@ -160,6 +143,19 @@ int main(int argc, char* argv[])
...
+    const char* sampler_type = "simple";
+    args.AddOption(&sampler_type, "--sampler-type", "--sampler-type",
+                   "Which sampler to use for coefficient: simple, pde");
+    double kappa = 0.001;
+    args.AddOption(&kappa, "--kappa", "--kappa",
+                   "Correlation length for Gaussian samples.");
+    int num_samples = 3;
+    args.AddOption(&num_samples, "--num-samples", "--num-samples",
+                   "Number of samples to draw and simulate.");
+    bool blank_perm = false;
```
gelever: would it be simpler to have the default empty filename string imply blank permutation?  That way all the complexity is moved into `IPF::ReadPermeabilityFile`. Everything else would be unchanged: there would be the single spe10 constructor the and the spe10 object would not need to be a unique pointer.

examples/mlmc.cpp
```diff
@@ -160,6 +143,19 @@ int main(int argc, char* argv[])
...
+    args.AddOption(&blank_perm, "--blank-perm", "--blank-perm",
+                   "--no-blank-perm", "--no-blank-perm",
+                   "Instead of loading SPE10 data, use constant permeability.");
     args.Parse();
     if (!args.Good())
     {
```
gelever: `MPI_Finalize();` is incorrect in this block.  The `mpi_session` from above takes care of it.

edit: Same w/ `examples/sampler.cpp`. Examples handle MPI inconsistently.  We should probably use `mpi_session` in all the examples and avoid manually `MPI_Finalize()`.

src/Sampler.hpp
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+       Pick a new sample; after calling this, GetFineCoefficient()
+       and GetCoarseCoefficient will return (versions of) the same
+       coefficient.
+    */
+    virtual void NewSample() {}
+
+    /// return current sample realized on fine mesh
+    virtual mfem::Vector& GetFineCoefficient() = 0;
```
gelever: This can return `const mfem::Vector&`.

src/Upscale.hpp
```diff
@@ -123,6 +123,12 @@ class Upscale : public mfem::Operator
     virtual MixedMatrix& GetCoarseMatrix();
     virtual const MixedMatrix& GetCoarseMatrix() const;
 
+    virtual const mfem::SparseMatrix& GetVertexProjection() const { return coarsener_->get_Pu(); }
```
gelever: Is this used anywhere?

src/Sampler.cpp
```diff
@@ -0,0 +1,233 @@
...
+            coefficient_coarse_(agg_index++) =
+                std::exp(coarse_sol(i) / coarse_constant_rep(i));
+        }
+    }
+    MFEM_ASSERT(agg_index == num_coarse_aggs_, "Something wrong in coarse_constant_rep!");
+
+    return coefficient_coarse_;
+}
+
+/// @todo ugly hack for debugging, @deprecated
```
gelever: Adding an already deprecated function in a PR doesn't sound like a good idea.
If it is still needed, maybe remove the tag.

src/utilities.hpp
```diff
@@ -421,6 +422,10 @@ void RescaleVector(const mfem::Vector& scaling, mfem::Vector& vec);
 */
 void GetElementColoring(mfem::Array<int>& colors, const mfem::SparseMatrix& el_el);
 
+void FVMeshCartesianPartition(
```
gelever, line 425: I know this was just copied from the earlier example files, but some of these can be made const.  We might as well take care of it in this PR.

.dir-locals.el
```diff
@@ -9,5 +9,7 @@
  (c-mode .
          ((c-basic-offset . 4)
           (c-indent-level . 4)
-          (c-file-style . "BSD"))))
+          (c-file-style . "BSD")))
```
gelever: Does this file even belong in the project?

src/Upscale.hpp
```diff
@@ -123,6 +123,12 @@ class Upscale : public mfem::Operator
...
+
+    /// @todo I feel like needing this reflects some leaky abstractions
+    ///       or bad design in one way or another.
+    virtual const GraphCoarsen& GetGraphCoarsen() const { return coarsener_->get_GraphCoarsen_ref(); }
```
gelever: Exposing this can be avoided using @chakshinglee's note on computing coarse constant using `P_vertex`.  Either in this class or outside utility function.
```
mfem::Vector GetCoarseConstant(const Upscale& upscale) const
{
         mfem::Vector ones = upscale.GetFineVector();
         ones = 1.0;

         return upscale.Restrict(ones);
}

**Review by gelever — APPROVED** [review]

File: src/Sampler.cpp [hunk 16]
```diff
@@ -0,0 +1,233 @@
...
+
+    double nu_parameter;
+    MFEM_ASSERT(dimension == 2 || dimension == 3, "Invalid dimension!");
+    if (dimension == 2)
+        nu_parameter = 1.0;
+    else
+        nu_parameter = 0.5;
+    double ddim = static_cast<double>(dimension);
+    scalar_g_ = std::pow(4.0 * M_PI, ddim / 4.0) * std::pow(kappa, nu_parameter) *
+                std::sqrt( tgamma(nu_parameter + ddim / 2.0) / tgamma(nu_parameter) );
```
gelever: `tgamma` should probably be `std::tgamma` to match `pow` and `sqrt`

src/Sampler.cpp
barker29: I removed the tag. I like what this does in `examples/sampler.cpp`, so I think it is worth keeping around.

examples/sampler.cpp
```diff
@@ -0,0 +1,405 @@
...
+        std::stringstream coarsenamev;
+        coarsenamev << "coarse_variance";
+        SaveFigure(mean_upscaled, ufespace, coarsenamev.str());
+        std::stringstream finenamev;
+        finenamev << "fine_variance";
+        SaveFigure(mean_fine, ufespace, finenamev.str());
+    }
+
+    if (myid == 0)
+        std::cout << picojson::value(serialize).serialize() << std::endl;
```
gelever: can this output w/ `picojson::value(serialize).serialize(true)` instead? It is easier to read when the output is vertically stacked.

src/Upscale.hpp
barker29: No. But I could swear I put it in there for some reason. Gone now.

src/GraphCoarsen.cpp
barker29: I feel fairly dumb for not realizing this myself.

src/Sampler.cpp
chakshinglee: Actually, the current method that you implement (if we exclude cell volume and boundary condition as well) can be interpreted as a graph-based sampler through graph Laplacian. Clearly it has its PDE origin in our case, but this extension may be used to apply to other problems (if we eventually encounter one some day Zzzz). 

Cell volume and other geometrical factors can be integrated into the graph through vertex and edge weights. So that seems to be not a big problem.

For boundary condition, apparently it has been an issue in the original PDE-based method, so it may be not too bad to use Neumann condition, which means simply ignore the boundary. In the reference paper, they solve problem on an extended domain, not sure how to do that in graph nicely if we remove the notion of boundary (the equivalent thing should be to only extend the graph at the vertices that is on the "boundary").

src/Sampler.cpp
```diff
@@ -0,0 +1,232 @@
...
+    fvupscale_->SolveCoarse(rhs_coarse_, coefficient_coarse_);
+    coefficient_coarse_ *= -1.0; // ??
+
+    const mfem::Vector& coarse_constant_rep = fvupscale_->GetCoarseConstantRep();
+    MFEM_ASSERT(coarse_constant_rep.Size() == coefficient_coarse_.Size(),
+                "PDESampler::GetCoarseCoefficient : Sizes do not match!");
+    for (int i = 0; i < coefficient_coarse_.Size(); ++i)
+    {
+        coefficient_coarse_(i) =
+            std::exp(coefficient_coarse_(i) / coarse_constant_rep(i)) * coarse_constant_rep(i);
```
gelever: This breaks if using more than just the constant basis vectors yeah?  Because `coarse_constant_rep[i]` is zero in some places and consequently divide by zero.
For example, `./sampler -m 4 -t 1.0 -vis` fails to run.

examples/sampler.cpp
```diff
@@ -0,0 +1,404 @@
...
+        SaveFigure(mean_upscaled, ufespace, coarsename.str());
+        std::stringstream finename;
+        finename << "fine_mean";
+        SaveFigure(mean_fine, ufespace, finename.str());
+        std::stringstream coarsenamev;
+        coarsenamev << "coarse_variance";
+        SaveFigure(mean_upscaled, ufespace, coarsenamev.str());
+        std::stringstream finenamev;
+        finenamev << "fine_variance";
+        SaveFigure(mean_fine, ufespace, finenamev.str());
```
gelever: `mean_upscaled` and `mean_fine` are both saved twice.  Is `coarse_variance` supposed to save `m2_upscaled` (and fine equivalent)?

These can be done a bit cleaner w/o stringstream
```c++
{
        SaveFigure(mean_upscaled, ufespace, "coarse_mean");
        SaveFigure(mean_fine, ufespace, "fine_mean");
        SaveFigure(m2_upscaled, ufespace, "coarse_variance");
        SaveFigure(m2_fine, ufespace, "fine_variance");
}
```

src/Sampler.cpp
```diff
@@ -0,0 +1,232 @@
...
+{
+    MFEM_ASSERT(current_state_ == FINE_SAMPLE ||
+                current_state_ == COARSE_SAMPLE,
+                "PDESampler object in wrong state (call NewSample() first)!");
+
+    if (current_state_ == FINE_SAMPLE)
+        fvupscale_->Restrict(rhs_fine_, rhs_coarse_);
+    coefficient_coarse_.SetSize(rhs_coarse_.Size());
+    fvupscale_->SolveCoarse(rhs_coarse_, coefficient_coarse_);
+    coefficient_coarse_ *= -1.0; // ??
```
gelever: I think this `*= -1.0;` needs to go into `Upscale::SolveCoarse` instead.  There may not  be anywhere else where it was used.

src/Upscale.cpp
```diff
@@ -104,6 +104,7 @@ void Upscale::SolveCoarse(const mfem::BlockVector& x, mfem::BlockVector& y) cons
 
     coarse_solver_->Solve(x, y);
     y *= -1.0;
+    OrthogonalizeCoarse(y);
```
gelever, line 109: This (lines 106 and 107) should probably be added to the non-block `SolveCoarse` also (right after line 90).

src/Upscale.cpp
```diff
@@ -244,6 +245,20 @@ void Upscale::Orthogonalize(mfem::BlockVector& vect) const
     Orthogonalize(vect.GetBlock(1));
 }
 
+void Upscale::OrthogonalizeCoarse(mfem::Vector& vect) const
+{
+    const mfem::Vector coarse_constant_rep = GetCoarseConstantRep();
+    double local_dot = (vect * coarse_constant_rep);
+    double global_dot;
+    MPI_Allreduce(&local_dot, &global_dot, 1, MPI_DOUBLE, MPI_SUM, comm_);
+    vect.Add(-global_dot, coarse_constant_rep);
```
gelever: @chakshinglee Referring to your note from an earlier commit, should this include the fine level global size somewhere to compute average?  Or is that automatically handled already?

`vect.Add(-global_dot / fine_global_size, coarse_constant_rep);`

chakshinglee: Good catch, you are right. I missed the inner product of constant vector. I actually missed that in the fine level formula, and consequently the coarse one as well (because it was derived from fine). It should be v -= (v * 1 ) / (1 * 1) * 1. Similarly for coarse level v_c -= (v_c * 1_c ) / (1_c * 1_c) * 1_c.

The inner product (1_c * 1_c) should have the same value as (1 * 1), which equals `fine_global_size`. Either include that factor in the normalization, or if we store this constant representation, we can just store the normalized version, i.e. store 1_c / || 1_c ||, then (1_c * 1_c) = 1 and we dont need that factor.

barker29:
A lot of comments here, so I want to write down what I think still needs to be done so you guys can check if I missed something:

- [x] PDESampler should handle real finite volumes, or at least document how to use the graph structure to do a finite volume problem.
- [x] Save the `coarse_one_rep_` vector instead of generating it every time.
- [x] Better handling of empty filename / blank permeability file in `mlmc.cpp`

Am I missing anything?

chakshinglee:
Just found a minor bug that is not from this PR, but maybe it can be corrected in this PR instead of open another one.

https://github.com/LLNL/smoothG/blob/b1b60725b4afc39b027b74d2734939a77612a4ab/src/HybridSolver.cpp#L906-L911

Line 908 should call `SetRelTol` instead of `SetMaxIter`.

PR #22, opened 2018-06-07
Robust ordering in SharedEntityCommunication broadcast [bugfix/broadcast-order]

barker29:
Changes to SharedEntityCommunication.hpp so that if two entities are both shared by the same processors, and have the same size, the ordering of their communication will be fixed and not depend on how they are ordered in entity_trueentity on each of the processors.

This change is due to a colleague who does not want a github account.

### DIFF
--- a/src/sharedentitycommunication.hpp
+++ b/src/sharedentitycommunication.hpp
@@ -44,6 +44,7 @@
 #include <sstream>
 #include <utility>
 #include <vector>
+#include <map>
 
 #include <mpi.h>
 #include <mfem.hpp>
@@ -124,6 +125,12 @@ public:
        Broadcast ints from master to slaves
        the idea here is if we know size in advance we can
        cut out a communication step
+
+       XXX: This function requires that the entities in ete_diag and
+            ete_offd be ordered consistently with the global true entity
+            ordering on all CPUs. This is the case with ete_diag, so the
+            requirement is reduced to ete_col_map being sorted in increasing
+            order.
     */
     void BroadcastFixedSize(int* values, int num_per_entity);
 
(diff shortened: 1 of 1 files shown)


gelever (approved):
Looks good to me.

Is there a simple test case where this would previously fail/ not work correctly?

src/sharedentitycommunication.hpp
```diff
@@ -124,6 +125,12 @@ class SharedEntityCommunication
        Broadcast ints from master to slaves
        the idea here is if we know size in advance we can
        cut out a communication step
+
+       XXX: This function requires that the entities in ete_diag and
+            ete_offd be ordered consistently with the global true entity
+            ordering on all CPUs. This is the case with ete_diag, so the
+            requirement is reduced to ete_col_map being sorted in increasing
+            order.
```
chakshinglee, line 133: Isn't this always the case for `HYPRE_ParCSRMatrix`?

src/sharedentitycommunication.hpp
```diff
@@ -644,18 +652,37 @@ void SharedEntityCommunication<T>::BroadcastSizes(T** mats)
...
 {
+    const int header_length = size_specifier_ + 1;
     int send_counter = 0;
     int receive_counter = 0;
     data_requests_ = new MPI_Request[num_master_comms_ + num_slave_comms_];
     MPI_Status* data_statuses =
         new MPI_Status[num_master_comms_ + num_slave_comms_];
 
+    // Invert the entity to true entity relation to obtain true entity to entity.
+    // This is the simplest thing to do and costs O(n log n), where n is the number
```
gelever, line 663: With `unordered_map` this cost can be dropped down to `O(n)`, though the difference in performance would probably be negligible.

src/sharedentitycommunication.hpp
barker29: That's the whole question. It seems to be true in practice, but is it guaranteed?

chakshinglee (approved):

src/sharedentitycommunication.hpp
chakshinglee: I asked Ruipeng about this, and he said if the matrix is created inside hypre (e.g. parmult or transpose), then col_map should be sorted, but if the matrix is created, say, through mfem wrapper, and the input col_map is not sorted, then hypre won't automatically sort it for us. This means it is not guaranteed.

PR #23, opened 2018-06-18
[Discussion] smoothG w/o MFEM

gelever:
This PR removes the dependency on MFEM and replaces it w/ independent linear algebra library.

There are a few code restructures through out but the core algorithms should be teh same.  More documentation can be found at [https://github.com/gelever/smoothG/pull/1](https://github.com/gelever/smoothG/pull/1).  I don't want to force any changes onto LLNL/smoothG, so if there are issues you don't like, please note them.

Some differences to note:
1. `MixedMatrix` uses dense element matrices throughout
1. The fine level is always solved using `SPDSolver`
    * CG preconditioned w/ BoomerAMG
        * assumes fine level `M` is diagonal
    * notable changes to operator complexity/ timing for tests/comparisons
1. Additional **TravisCI** environments:
    * linux-gcc, linux-clang, osx-gcc, osx-clang
1. Additional build scripts
    * `INSTALL.md` not updated yet, as I expect linear algebra library integration to change
1. No **SAAMGE** support as of yet
1. No support yet for various trace types
1. Finite volume examples are not generated on the fly, the graph information is simply loaded from file into `generalgraph`.
    * finite volume graph data accounts for most of the `206,789 additions`
    * I'm not sure how this will work w/ well models, may require significant changes (?)
    * there exists a separate repo that integrates MFEM and gelever/smoothG as PoC: https://github.com/gelever/ReservoirSimulator2000
1. Assumes input is well defined graph with each edge connected to 2 vertices
    * allows eliminating edge dofs when solving to be compatibale w/ **MFEM** generated problems
        * not very elegant and may require changes
1. Argument parsing is very minimal and just enough to handle what we had in examples
   * Previously provided by **MFEM**
   * This can be improved upon or replaced entirely with external library
1. More modern **CMake** scripts
1. Implementation and header files are separated into `src` and `include`
1. linear algebra libraries are not as thoroughly tested as **MFEM** so there may still be bugs
1. git log should probably be squashed down significantly
1. And so on and so on


src/Utilities.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+    SparseMatrix offd(std::move(offd_indptr), std::move(offd_indices), std::move(offd_data),
+                      num_edges_local, num_edge_diff);
+
+    return ParMatrix(comm, starts[0], starts[1],
+                     std::move(diag), std::move(offd),
+                     std::move(col_map));
+}
+
+SparseMatrix RestrictInterior(const SparseMatrix& mat)
```
chakshinglee, line 144: Based on this function itself, I think it is not so intuitive why it is called `RestrictInterior`. I understand this comes from the `GraphTopology`, but since it is taken out of the class as an independent function, I think we can call it something like 
```
SparseMatrix RemoveLargeEntries(const SparseMatrix& mat, double val)  
```
Here `val` is the threshold. And possibly it may be used somewhere else later.

include/Graph.hpp
```diff
@@ -0,0 +1,114 @@
...
+    std::vector<int> part_local_;
+
+    // Graph relationships
+    SparseMatrix vertex_edge_local_;
+    ParMatrix edge_true_edge_;
+    ParMatrix edge_edge_;
+
+    // I'm not sure this is `graph` information?
+    // Graph Weight / Block
+    std::vector<double> weight_local_;
```
chakshinglee, line 100: I think if we associate `weight_local_` and `W_local_` with edge weight and vertex weight of the graph respectively, then they should be parts of the graph. So I prefer to name them accordingly. `weight_local_` is originally edge weight in the usual sense, so there is not doubt about it. `W_local_` is a bit awkward, I am not sure. Especially when `W_local_` is not diagonal, it should probably not belong here. So far have we had any actual application where `W_local_` is not diagonal? If yes, then I feel like it should be moved to MixedMatrix; otherwise I think we can call it vertex weight and store it as a vector for now.

src/GraphTopology.cpp
```diff
@@ -13,469 +13,301 @@
...
+    ParMatrix edge_face = face_edge_.Transpose();
 
-    // nfaces_bdr = number of global boundary faces in this processor
-    int nfaces_bdr = 0;
-    mfem::SparseMatrix aggregate_boundaryattr;
-    if (edge_boundaryattr)
-    {
-        mfem::SparseMatrix tmp = smoothg::Mult(aggregate_edge, *edge_boundaryattr);
-        aggregate_boundaryattr.Swap(tmp);
+    face_face_ = parlinalgcpp::RAP(edge_edge, edge_face);
```
chakshinglee, line 98: I am thinking if it worths to make a function for triple matrix product `Matrix Mult(Matrix&, Matrix&, Matrix&)`. I dont know about the new code, but I remember there were several places in the old code where we are doing this kind of RAP, but since the specific assumption of RAP function, we end up doing the transpose step twice. I dont think this extra step of transpose affects the performance that much, but just feel like wasted.

Another way could be to make a function call `PAR` (sounds confusing though), which transpose the other side.

src/GraphTopology.cpp
```diff
@@ -13,469 +13,301 @@
...
-    // Need to sort the edge indices to prevent index problem in face_edge
-    aggregate_edge.SortColumnIndices();
+    ParMatrix edge_agg_ext = edge_edge.Mult(edge_agg_d);
+    ParMatrix agg_agg = agg_edge_d.Mult(edge_agg_ext);
 
-    AggregateEdge2AggregateEdgeInt(aggregate_edge, Agg_edge_);
-    mfem::SparseMatrix edge_aggregate(smoothg::Transpose(aggregate_edge));
+    agg_edge_ext = 1.0;
+    SparseMatrix face_agg_int = MakeFaceAggInt(agg_agg);
+    SparseMatrix face_edge_ext = face_agg_int.Mult(agg_edge_ext);
```
chakshinglee, line 85: I prefer to call `face_agg_int` either `intface_agg` or using the convention in the new code `face_int_agg` because int here refers to interior face.

And for `face_edge_ext`, I guess it comes from `face_agg * agg_edge_ext` and the middle is eliminated. There are other places using this naming convention. What is the reason you dont like the original name? The original name in the old code is `intface_Agg_edge` because the algorithm comes from Sec 2.4 of Panayot's paper (doi:10.1002/nla.300), and in that paper, the name is "AEface_AE_face". I think it is easier for other people to follow if we use some what equivalent names (say `face_agg_edge`).

src/GraphTopology.cpp
```diff
@@ -13,469 +13,301 @@
...
 
-    // Build face "dof-true dof-dof" table from local face_edge and
-    // the edge "dof-true dof-dof" table
-    GenerateOffsets(comm_, nfaces, face_start_);
+SparseMatrix GraphTopology::MakeFaceEdge(const ParMatrix& agg_agg,
+                                         const ParMatrix& edge_ext_agg,
+                                         const SparseMatrix& agg_edge_ext,
+                                         const SparseMatrix& face_edge_ext)
+{
+    const auto& agg_agg_diag = agg_agg.GetDiag();
```
chakshinglee, line 205: I think `agg_agg_diag` is not really needed other than defining `num_aggs`, which can be obtained by `agg_agg_offd.Rows()`.

include/GraphUpscale.hpp
```diff
@@ -0,0 +1,318 @@
...
+       @brief Graph Constructor
+
+       @param graph contains input graph information
+       @param spect_tol spectral tolerance determines how many eigenvectors to
+                        keep per aggregate
+       @param max_evects maximum number of eigenvectors to keep per aggregate
+       @param hybridization use hybridization as solver
+    */
+    GraphUpscale(Graph graph, double spect_tol = 0.001, int max_evects = 4,
+                 bool hybridization = false, const std::vector<int>& elim_edge_dofs = {});
```
chakshinglee, line 61: Currently there is no example that uses `elim_edge_dof`, right? Since `GraphTopology` does not produce "coarse boundary", we need to be careful. If I understand correctly, the boundary is identified by partition. This can produce some unexpected results if the user is not aware of the assumption. 

I guess we need to think about how to handle different essential dofs (there will be cases where vertex dofs is fixed). But let's do this in a separate PR. For this PR, am I right that `elim_edge_dofs` is always empty?

include/Graph.hpp
gelever: So far `W` block has been diagonal when we have used it.  It is convenient to keep it here because all the problem info is stored together.

src/Utilities.cpp
gelever: k, I'll rename it

src/GraphTopology.cpp
gelever: Yes this would be a helpful function.  Calling it `Mult` would probably be most clear since that is exactly what it will be doing, without any transpose.

src/GraphTopology.cpp
gelever: No specific reason for the names,  I'll modify these to make them consistent/more clear.

include/GraphUpscale.hpp
gelever: Correct, so far the included examples don't use this.

src/MixedMatrix.cpp
```diff
@@ -15,221 +15,311 @@
...
-    assert(edge_d_td);
-    assert(weight.Size() == vertex_edge.Width());
-    assert(edge_d_td_->Height() == vertex_edge.Width());
+    int num_vertices = D_local_.Rows();
+    int num_edges = D_local_.Cols();
+
+    M_elem_.resize(num_vertices);
 
-    mfem::Vector weight_cut;
+    SparseMatrix edge_vertex = D_local_.Transpose();
```
chakshinglee, line 39: `edge_vertex` is unused?

gelever: Good catch.

src/MixedMatrix.cpp
```diff
@@ -15,221 +15,311 @@
...
-                         DistributeWeight dist_weight)
-    : edge_d_td_(&edge_d_td),
-      edge_td_d_(edge_d_td_->Transpose())
+MixedMatrix::MixedMatrix(const Graph& graph)
+    : edge_true_edge_(graph.edge_true_edge_),
+      D_local_(MakeLocalD(graph.edge_true_edge_, graph.vertex_edge_local_)),
+      W_local_(graph.W_local_),
+      elem_dof_(graph.vertex_edge_local_),
+      agg_vertexdof_(SparseIdentity(D_local_.Rows())),
+      face_facedof_(SparseIdentity(elem_dof_.Cols()))
```
chakshinglee, line 32: I guess `agg_vertexdof_` and `face_facedof_` come from the old code (maybe `HybridSolver`). They are pretty bad names (I think I am the one who gave them these names) in the sense that there is no `agg` nor `face` in a certain level. They are `vertex_vdof` `edge_edof` in the multilevel code.

`elem_dof` is a bit trickier, it should be called `vertex_edof` if we follow the same convention above. But this will have a naming conflict with the ml code (`vertex_edof` means something else in ml code), so lets leave it as it is for now. But I think we need to do something to it later.

src/MixedMatrix.cpp
```diff
@@ -15,221 +15,311 @@
...
+    M_elem_.resize(num_vertices);
 
-    mfem::Vector weight_cut;
+    SparseMatrix edge_vertex = D_local_.Transpose();
+    std::vector<double> weight_inv = graph.weight_local_;
+
+    for (auto& i : weight_inv)
+    {
+        assert(std::fabs(i) > 1e-12);
+        assert(i == i);
```
chakshinglee, line 45: This was something else I believe, but certainly we dont need it now XD

src/MixedMatrix.cpp
gelever: Yes, there are a bunch more relationships from the ML code that can be aggregated into one structure later on.  You had `GraphSpace` before, or something like that.

src/MixedMatrix.cpp
gelever: This was to make sure is not `NaN`, though i will change to `std::isfinite` for a better check.

src/MixedMatrix.cpp
```diff
@@ -15,221 +15,311 @@
...
+
+    MPI_Allreduce(&local_size, &global_size, 1, MPI_INT, MPI_MAX, D_global_.GetComm());
+
+    const double zero_tol = 1e-6;
+
+    return global_size > 0 && W_global_.MaxNorm() > zero_tol;
+}
+
+
+ParMatrix MixedMatrix::ToPrimal() const
```
chakshinglee, line 212: Currently this is function is only called in `generalgraph.cpp` (correct me if I missed something). I am thinking to use this function also in `MinresBlockSolver` and `SPDSolver` because in theory they need the same operation, some adjustment is needed though.

src/MixedMatrix.cpp
```diff
@@ -15,221 +15,311 @@
...
     }
 
-    // Nonzero row of edge_owned means the edge is owned by the local proc
-    mfem::SparseMatrix edge_owned;
-    edge_d_td.GetDiag(edge_owned);
+    return A;
+}
 
-    mfem::SparseMatrix graphDT(smoothg::Transpose(vertex_edge));
+void MixedMatrix::AssembleM()
```
chakshinglee, line 230: This can simply call `AssembleM(const std::vector<double>& agg_weight)`? I understand this way will need to create an unnecessary const vector of 1s, but if we are willing to store a diagonal element matrix as dense matrix, this is extra vector of 1s won't really affect the performance? Especially I believe we dont need this function in loops (Monto Carlo or nonlinear solve).

include/GraphGenerator.hpp
```diff
@@ -44,27 +42,20 @@ class GraphGenerator
...
-       @param nvertices number of vertices of the graph to be generated
-       @param mean_degree average vertex degree of the generated graph
-       @param beta probability of rewiring
-       @param seed seed for the random number engine
-    */
-    GraphGenerator(int nvertices, int mean_degree, double beta, unsigned int seed);
+        @brief Generate a random graph based on the Watts-Strogatz model
 
+        Generate a ring of nvertices_ vertices, each vertex is connected
+        to the neighboring mean_degree_ vertices (mean_degree/2 on each side)
```
chakshinglee, line 53: I think this description is not correct? This is the initial step, but the final graph is not a ring.

src/GraphEdgeSolver.cpp
```diff
@@ -0,0 +1,195 @@
...
+}
+
+void GraphEdgeSolver::OffsetMult(int offset, const DenseMatrix& input, DenseMatrix& sigma_sol,
+                                 DenseMatrix& u_sol) const
+{
+    OffsetMult(offset, input.Cols(), input, sigma_sol, u_sol);
+}
+
+void GraphEdgeSolver::OffsetMult(int start, int end, const DenseMatrix& input,
+                                 DenseMatrix& output) const
```
chakshinglee, line 156: Can we collapse all these `OffsetMult` and `Mult` to one `Mult` function as far as offset is concerned? Say, put `start` and `end` to the end of the input list, and set default value for `start` to be `0`. I doubt we have any use case of `end != input.Cols()` currently (actually I only found one place where `OffsetMult` is called while there are 4 such functions)

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+      D_trace_sum_(gt_.agg_ext_edge_.Rows()),
+      D_trace_(gt_.agg_ext_edge_.Rows()),
+      F_potential_(gt_.agg_ext_edge_.Rows())
+{
+    ParMatrix permute_v = MakeExtPermutation(gt_.agg_ext_vertex_);
+    ParMatrix permute_e = MakeExtPermutation(gt_.agg_ext_edge_);
 
-  If we set it up properly we can have known input/output for
-  most of the functions in this class, I think...
+    ParMatrix permute_v_T = permute_v.Transpose();
```
chakshinglee, line 41: This is not needed?

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+        D_sub_T.InverseScaleRows(M_sub);
+
+        SparseMatrix DMinvDT = D_sub.Mult(D_sub_T);
+
+        eigs.Compute(DMinvDT, evects);
+
+        if (evects.Cols() > 1)
+        {
+            DenseMatrix evects_ortho = evects.GetCol(1, evects.Cols());
+            agg_ext_sigma_[agg] = D_sub_T.Mult(evects_ortho);
```
chakshinglee, line 153: This can reduce to one line (and avoid copying) if you can add something like the idea of `OffsetMult` in `GraphEdgeSolver` to `DenseMatrix`, but it's up to you (only if you think this kind of function will in useful in general for the linear algebra library).

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+        else
+        {
+            agg_ext_sigma_[agg].SetSize(D_sub_T.Rows(), 0);
+        }
+
+        DenseMatrix evects_restricted = RestrictLocal(evects, col_marker_,
+                                                      vertex_dofs_ext, vertex_dofs_local);
+
+        VectorView first_vect = evects_restricted.GetColView(0);
+        vertex_targets_[agg] = smoothg::Orthogonalize(evects_restricted, first_vect, 1, max_evects_);
```
chakshinglee, line 164: is the scope `smoothg` necessary here?

src/MixedMatrix.cpp
chakshinglee: Aw, I see, now it makes sense to me. Thanks for the explanation, I was wondering why there is this check.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+        const auto& face_M = shared_M[face];
+        const auto& face_D = shared_D[face];
+        const auto& face_sigma = shared_sigma[face];
+
+        linalgcpp::HStack(face_sigma, collected_sigma);
+
+        bool shared = face_shared.RowSize(face) > 0;
+
+        // TODO(gelever1): resolve this copy.  (Types must match so face_? gets copied and promoted
+        // to rvalue).
```
chakshinglee, line 310: Does RVO apply to this case?

include/GraphCoarsen.hpp
```diff
@@ -0,0 +1,208 @@
...
+    double spect_tol_;
+
+    SparseMatrix P_edge_;
+    SparseMatrix P_vertex_;
+    SparseMatrix face_cdof_;
+    SparseMatrix agg_bubble_dof_;
+
+    std::vector<DenseMatrix> vertex_targets_;
+    std::vector<DenseMatrix> edge_targets_;
+    std::vector<DenseMatrix> agg_ext_sigma_;
```
chakshinglee, line 195: These three were temporary objects before. Are you thinking about preparing for adaptive code? Otherwise at least their allocated memory can be destroyed after the coarsening process? 

I think most likely `agg_ext_sigma_` do not need to be kept even in adaptive algorithm?

src/Graph.cpp
```diff
@@ -0,0 +1,225 @@
...
+            assert(std::fabs(global_weight[edge_map_[i]]) > 1e-14);
+            weight_local_[i] = std::fabs(global_weight[edge_map_[i]]);
+        }
+    }
+    else
+    {
+        std::fill(std::begin(weight_local_), std::end(weight_local_), 1.0);
+    }
+
+    const SparseMatrix& edge_offd = edge_edge_.GetOffd();
```
chakshinglee, line 99: Some suggestions in mind (probably in a separate PR): 
Currently `edge_edge_` is used for two purposes: 1. check if an edge is shared; 2. construct `GraphTopology`.

Since now we assume no weird graphs (edges associated with only one vertex), I think we can determine if an edge is shared by simply checking the row size of "edge_vertex_local"? This way we dont need to store `edge_edge_` in `Graph`. We may store edge_vertex instead of vertex_edge in `Graph`. The former probably requires less memory, but if we dont need to store the extra `edge_edge_` variable, then overall memory consumption is lower. 

`edge_edge_` is still needed in `GraphTopology`, but it can be used as a temporary object.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-        GetTableRow(Agg_edge, i, local_fine_dofs);
-        GetTableRow(Agg_face, i, faces);
-        nlocal_coarse_dofs = vertex_target[i].Width() - 1;
-        for (int j = 0; j < faces.Size(); ++j)
-            nlocal_coarse_dofs += edge_traces[faces[j]].Width();
-        for (int j = 0; j < local_fine_dofs.Size(); ++j)
-            Pedges_i[local_fine_dofs[j] + 1] = nlocal_coarse_dofs;
-        if (build_coarse_relation)
+        DenseMatrix& edge_traces(edge_targets_[face]);
+        if (edge_traces.Cols() < 1)
```
chakshinglee, line 336: We need to always have `edge_traces.Cols() >= 1`?

include/GraphCoarsen.hpp
```diff
@@ -0,0 +1,208 @@
...
+    BlockVector Restrict(const BlockVector& fine_vect) const;
+
+    /** @brief Restrict a fine level mixed form vector to the coarse level
+        @param fine_vect fine level mixed form vector
+        @param coarse_vect restricted mixed form vector
+    */
+    void Restrict(const BlockVector& fine_vect, BlockVector& coarse_vect) const;
+
+    /** @brief Get the face to coarse dof relationship */
+    const SparseMatrix& GetFaceCDof() const { return face_cdof_; }
```
chakshinglee, line 130: This function may be deprecated.

chakshinglee: (because the relation is store in `MixedMatrix` now)

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-
-    return unique_ptr<mfem::HypreParMatrix>(d_td);
+
+    ParMatrix edge_true_edge = BuildEdgeTrueEdge();
+
+    return MixedMatrix(std::move(M_elem), std::move(agg_cdof_edge),
+                       std::move(D_c), std::move(W_c),
+                       std::move(edge_true_edge),
+                       BuildAggCDofVertex(),
+                       face_cdof_);
```
chakshinglee, line 1056: Is it right that the last two matrices are copied in the constructor of `MixedMatrix`? For `face_cdof_`, I guess the reason is because it is a member in `GraphCoarsen` and there is a function to reference it. If we decided to not keep it in `GraphCoarsen`, can we also std::move it?

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-        edge_cd_start_, &d_td_d_diag, &d_td_d_offd, d_td_d_map);
-
-    // Create a selection matrix to set dofs on true faces to be true dofs
-    int* select_i = new int[ncdofs + 1];
-    select_i[0] = 0;
-    int cdof_counter = 0;
-    for (int i = 0; i < nfaces; i++)
+
+        h_D_trace.MultAT(h_F_potential, DTT_F);
+        h_D_trace.MultAT(B_potential_[agg], DTT_B);
```
chakshinglee, line 923: Can we implement (potential * D) * trace instead of potential * (D * trace), that is, computing (potential * D) first. The reason is because (I think) the former has higher chance to be faster. This depends on whether potential or trace is longer in length (which in turn depends on the graph and partition). At least for graphs coming from conforming meshes, I think trace is always shorter. 

This is the reason why the original code computes (potential * D) first. That said, I think the timing difference will be outweighed by the solving time of local eigen problems, so I am not too worried about it, but I just think if possible, why not. I guess the drawback would be, the code will be not as clean as the current version. The main problem is traces are not of the same length, so here will need an extra loop over faces. I just want to bring this up and you can think about it.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+    for (int i = 0; i < fine_size; ++i)
     {
-        unique_ptr<mfem::SparseMatrix> W_tmp(mfem::RAP(Pvertices, *W_proc_, Pvertices));
-        mfem::SparseMatrix W_thresh(*W_tmp);
+        int edge_dof = edge_dofs_on_face[i];
+        auto verts = edge_vertex.GetIndices(edge_dof);
+        int vert = (partition[verts[0]] == agg ) ? verts[0] : verts[1];
+
+        std::vector<int> fine_edges = mgl.GetElemDof().GetIndices(vert);
+        SetMarker(col_marker, fine_edges);
```
chakshinglee, line 876: It seems to me that complexity of the original implementation (using `break`) is lower, isn't it?

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
     }
 
-    CoarseD_->Finalize();
+    std::vector<double> data(indices.size(), 1.0);
+
+    return SparseMatrix(std::move(indptr), std::move(indices), std::move(data),
+                        num_aggs, num_cdofs);
+}
+
+ParMatrix GraphCoarsen::BuildEdgeTrueEdge() const
```
chakshinglee, line 733: This name is misleading - its about coarse edge dofs, not edge.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-    if (build_coarse_relation)
+    for (int i = 0; i < num_faces; ++i)
     {
-        edge_cdof_marker2.SetSize(Agg_cdof_edge_->Width());
-        edge_cdof_marker2 = -1;
+        if (ftf_diag.RowSize(i) > 0)
+        {
+            num_true_dofs += face_cdof_.RowSize(i);
+        }
     }
```
chakshinglee, line 749: I think `num_true_dofs` is not really used in the rest of the function, and therefore this loop is not needed. Consequently, `ftf_diag` and `num_traces` are also not needed? Or you were using them for other purpose?

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-                Pedges_j[ptr] = facecdofs[k];
-                // since we did not do local_rhs *= -1, we store -trace here
-                Pedges_data[ptr++] = -edge_traces_i(j, k);
-            }
+            offd_nnz++;
         }
 
-        // store global and local coarse M
-        if (build_coarse_relation)
+        offd_indptr[i + 1] = offd_nnz;
```
chakshinglee, line 776: Just curious, is there some reason to change the code of filling `offd_indptr`. To me, the new code seems doing the exact same thing.

src/GraphCoarsen.cpp
chakshinglee: I guess some of the variables inside this function as well, e.g. `cface_cface`.

include/GraphGenerator.hpp
gelever: I'll add the rewiring portion to the description.

src/GraphEdgeSolver.cpp
gelever: Yes, this can be simplified into fewer functions.

src/GraphCoarsen.cpp
gelever: Yes, `OffsetMult` can be added to `Operator` class since it doesn't really matter what is applying the `Mult`.

src/GraphCoarsen.cpp
gelever: At one point there was a similarly named function in the class, but it is now no longer necessary. :+1:

src/GraphCoarsen.cpp
gelever: I'm not sure.  The copy occurs because `face_M[0]` must be the same type as `Combine(face_M, num_faces_edges)` so `face_M[0]` is promoted to rvalue by copying.  I haven't looked too deeply into this, just noted that it forces a copy here.

src/GraphCoarsen.cpp
gelever: Yes, this should probably assert the condition instead of continuing.

src/GraphCoarsen.cpp
gelever: This was probably just poorly translated from the original master.  I'll remove all the unused parts.

src/GraphCoarsen.cpp
gelever: No reason, it should be doing the same thing.

src/GraphCoarsen.cpp
gelever: Yes, this may be faster.  Right now the ML code uses local RAP anyways, so I will have to revisit this idea when reimplementing assembling local `M` w/ potentials in the ML case.

src/GraphCoarsen.cpp
gelever: Only `face_cdof_` is copied since it is a member, the `BuildAggCDofVertex` returns an rvalue and is moved into the constructor.  It should be possible to move `face_cdof_` too since it is not used after coarsening.

include/GraphCoarsen.hpp
gelever: Correct.

include/GraphCoarsen.hpp
gelever: It is just easier to keep these as member functions right now to avoid passing them around to every function.  It may be useful to clear them after coarsening is complete.

src/MixedMatrix.cpp
gelever: Yeah the unweighted assemble is only called once during the initial setup phase.

src/GraphUpscale.cpp
```diff
@@ -15,215 +15,659 @@
...
+    setup_time_ += timer.TotalTime();
+}
+
+void GraphUpscale::MakeCoarseSolver()
+{
+    auto& mm = GetCoarseMatrix();
+
+    if (hybridization_)
+    {
+        coarse_solver_ = make_unique<HybridSolver>(mm);
```
chakshinglee, line 63: `HybridSolver` doesn't take fine/coarse elim_dofs, I guess currently this works because elim_dofs is always empty in all the examples. My concern here is, internally `HybridSolver` also does not consider elim_dofs, which is inconsistent with the other two solvers. At this point I am not sure if we should have elim_dofs (for edge dofs) at all, but at least the solvers should be consistent. If it's not too much work, I prefer to add back the notions of elim_dofs in `HybridSolver`, otherwise at least add a todo notification somewhere in the class to tell people this does not work when there is essential edge dofs.

chakshinglee: One difficulty I can think of is there no boundary indicator anymore, so you cannot find out which are the essential multiplier dofs, is it right?

gelever: At this point `elim_dofs` are only used to on the fine level solver (`SPDSolver`) when the input graph comes from MFEM.  I agree the solvers should have the same interface so I'll look into providing essential edge dofs for `HybridSolver`.

src/HybridSolver.cpp
```diff
@@ -20,673 +20,275 @@
...
-    mfem::SparseMatrix tmp = SparseIdentity(nvertices);
-    Agg_vertexdof_.Swap(tmp);
+    ParMatrix edge_td_d = mgl.EdgeTrueEdge().Transpose();
+    ParMatrix edge_edge = mgl.EdgeTrueEdge().Mult(edge_td_d);
+    ParMatrix edgedof_multiplier_d(comm_, std::move(edgedof_multiplier));
+    ParMatrix multiplier_d_td_d = parlinalgcpp::RAP(edge_edge, edgedof_multiplier_d);
 
-    Agg_edgedof_.MakeRef(D_);
-    const mfem::SparseMatrix edge_edgedof;
+    multiplier_d_td_ = MakeEntityTrueEntity(multiplier_d_td_d);
```
chakshinglee, line 51: Some suggestion (again probably another PR): line 46-51 here can group into one function? Also, seems computing `trueedgedof_multiplier` = `edge_td_d` * `edgedof_multiplier_d` and then multiply its transpose to it is a faster way to get `multiplier_d_td_d`? Again, the saving should be minimal.

src/HybridSolver.cpp
```diff
@@ -20,673 +20,275 @@
...
-        for (int i = 0; i < nlocal_edgedof; ++i)
-            edgedof_global_to_local_map[local_edgedof[i]] = -1;
+        AinvDMinvCT_i.SetSize(nlocal_vertexdof, nlocal_multiplier);
+        hybrid_elem.SetSize(nlocal_multiplier, nlocal_multiplier);
+        Aloc.SetSize(nlocal_vertexdof, nlocal_vertexdof);
+        DMinvCT.SetSize(nlocal_vertexdof, nlocal_multiplier);
 
-        MinvDT_i.InvLeftScaling(M_diag);
+        MinvDT_i.SetSize(Minv.Cols(), Dloc.Rows());
+        MinvCT_i.SetSize(Minv.Cols(), Cloc.Rows());
```
chakshinglee, line 223: Did we talk about (maybe you bring this up before? I don't remember) if we should set the right size of returned matrix inside various "`Mult`" functions. To me the only reason to do this outside of the `Mult` function is to provide one more debug check, and I don't think it is that necessary (maybe later if I got stuck by some strange bug then I will change my mind). What do you think?

src/HybridSolver.cpp
```diff
@@ -20,673 +20,275 @@
...
-        if (W_)
-        {
-            auto Wloc = ExtractRowAndColumns(*W_, local_vertexdof, local_vertexdof,
-                                             edgedof_global_to_local_map);
-            mfem::DenseMatrix tmpW;
-            Full(Wloc, tmpW);
+        Ainv_i.Mult(DMinvCT, AinvDMinvCT_i);
 
-            Aloc -= tmpW;
+        if (DMinvCT.Rows() > 0 && DMinvCT.Rows() > 0)
```
chakshinglee, line 244: One of them should be `Cols()`.  

On an unrelated issue, `DMinvCT.Cols() == 0` when number of local multipliers is 0, which happens when there is an isolated vertex. But I cannot think of any case when `DMinvCT.Rows() == 0` (number of local vertex dofs is 0). I feel like this should not be happening, so an assert maybe more appropriate. Am I missing some edge case?

src/HybridSolver.cpp
```diff
@@ -786,49 +378,82 @@ void HybridSolver::RecoverOriginalSolution(const mfem::Vector& HybridSol,
             HybridSol.GetSubVector(local_multiplier, mu_loc);
 
             // Compute u = (DMinvDT)^-1(f-DMinvC^T mu)
-            AinvDMinvCT_[iAgg].AddMult_a(-1., mu_loc, u_loc);
+            //AinvDMinvCT_[iAgg].AddMult_a(-1., mu_loc, u_loc);
+            tmp.SetSize(u_loc.size());
+            AinvDMinvCT_[iAgg].Mult(mu_loc, tmp);
+            u_loc -= tmp;
```
chakshinglee, line 384: There is no `AddMult` - equivalent function in `linalgcpp`? You are not a big fan of this function?

include/Utilities.hpp
```diff
@@ -0,0 +1,398 @@
...
+/** @brief Extract a dense submatrix from a sparse matrix
+    @param A matrix from which to extract
+    @param row row indices to extract
+    @param col column indices to extract
+    @param colMapper map of global to local indices
+    @param A_sub holds the extracted submatrix
+*/
+void ExtractSubMatrix(const SparseMatrix& A, const std::vector<int>& rows,
+                      const std::vector<int>& cols, const std::vector<int>& colMapper,
+                      DenseMatrix& A_sub);
```
chakshinglee, line 254: We don't need this function anymore?

include/Utilities.hpp
```diff
@@ -0,0 +1,398 @@
...
+
+/** @brief Adds two sparse matrices C = alpha * A + beta * B
+
+    @param alpha scale for A
+    @param A A matrix
+    @param beta scale for B
+    @param B B matrix
+    @returns C such that C = alpha * A + beta * B
+*/
+SparseMatrix Add(double alpha, const SparseMatrix& A, double beta, const SparseMatrix& B);
```
chakshinglee, line 280: These three functions, `Add`, `AssembleElemMat` and `MultScalarVVt`, are also not used anywhere in the library. Some of them may be moved to the linear algebra library or just really not needed anymore? Should we remove/deprecate them?

src/HybridSolver.cpp
gelever: Yeah I mentioned this earlier https://github.com/gelever/linalgcpp/issues/20
The only drawback of setting size automatically is that it does it silently and that may not be the size you expected.  Despite this i think it is worth doing, as it simplifies the code quite a bit.

src/HybridSolver.cpp
gelever: I have no qualms with `AddMult`, but it was not yet implemented so I have just done it by hand for now.  It may require some work since all the different operators may need to be rewritten to apply `AddMult` instead of `Mult` and then redefine `Mult` in terms of `AddMult`.

include/Utilities.hpp
gelever: Yes, if they aren't being used I'll remove them.

chakshinglee:
@gelever I like this PR and I think overall it is a very good piece of work, thanks a lot for your hard work! This new version definitely look much cleaner than the current master (I hope you understand that I do think restructuring the original code is desirable, I just think it would be easier to have a separate PR specifically for that purpose. Getting familiar with your new linear algebra code and the restructured smoothG at the same time is quite overwhelming for me). 

I did not go through `SharedEntityComm`, I originally just copied the same files from `ParElag` and add the case of `SparseMatrix`, so I don't think I know enough about the class to review it. I'm sure @barker29 can do a much better job.

Other than the individual comments in the files, I noticed that there are a few things missing:
1. Rescaling feature in `HybridSolver`, cf. member function `ComputeScaledHybridSystem` in the current master. 
2. Setting isolated vertices in the partitioner.
3. use edge weight in the partition algorithm.

Since this PR is to remove dependency on MFEM, we expect the test results to be identical/very close to the old ones. For comparison purpose, I do want to keep the original solver setting, and see how different are the results. That is, for now let's also use the minres/hybrid solver in the fine level. In fact, other than OC, I noticed upscaling errors in some of the tests are also different, e.g. fv-metis, poweriter, do you know why? I think in some of our conversations I agreed to use `SPDSolver` in the fine level, maybe I was not paying attention, but I did not mean to put it in this PR. I do think it is fairer to use the best solver (`SPDSolver`) in the fine level, so we should switch to it eventually. You can keep the `SPDSolver` files, but just call other solvers in `GraphUpscale`.

Lastly, I have expressed my opinion on this before, so just put them down again for the record. I still think at this point, the partition library (which you told me is basically a wrapper of metis) should live inside smoothG. There is some thought on better organization of the codes within our group (I am not sure it will eventually be implemented), if we will really do that, that will probably be a better time to take the partition code out. The initial thought includes hybridization and the SharedEntityComm code, as they appear in several libraries. Also, what is the reason that you don't put `SparseSolver` inside `linalgcpp`? Is it because you dont like the idea of optional dependency? In any case, maybe you can tell me what is your thought/plan on these two dependencies. Thanks.

src/HybridSolver.cpp
chakshinglee: Good, then we agree on this issue. 

In fact, matrix multiplication (say, A * B) makes sense as long as number of columns of A equals number of rows of B. Therefore, whether the resulting matrix has the intended size should not be the concern of the multiplication function. Instead, the size should be checked by the user (outside the function).

gelever:
@chakshinglee Thanks for taking the time to review this!

* `SharedEntityComm` should be very similar to the original, just modified to use the new linalg library.
* Partitioning can be moved into **smoothG** and then I can add the different partition options.
* I missed the `ComputeScaledHybridSystem` because I assumed it was part of SAAMGE code, but it seems to be indpedent, so I will add it.

`par/poweriter` tests are different errors because the solve tolerance is more lenient (faster testing).
I'm not sure why `fv-metis` is different, probably because different settings for metis between examples.  For other solvers on the fine level, if feels like a step backward to revert to other solvers, but the change should not be difficult.

For dependencies, I preferred modular design over optional dependencies.  If you want to rearrange the linear algebra libraries or include parts of them into **smoothG**, it should not be too difficult.

include/Utilities.hpp
gelever: Looks like `Add` is used in the sparse eigensolver for `A_minus_sigma_B_inv_`, although this function should probably just be moved into `linalgcpp`.

chakshinglee: My bad, I was using IDE to check these functions' usage. You can decide where to put this function.

include/GraphEdgeSolver.hpp
```diff
@@ -0,0 +1,205 @@
...
+    */
+    Vector Mult(const VectorView& rhs) const;
+
+    /**
+       @brief Solves \f$ (D M^{-1} D^T) u = g\f$, \f$ \sigma = M^{-1} D^T u \f$.
+
+       @param rhs \f$ g \f$ in the formula above
+       @param sigma_sol \f$ \sigma \f$ in the formula above
+    */
+    void Mult(const VectorView& rhs, VectorView output) const;
```
chakshinglee, line 108: better match the documentation and the function, either use replace `output` by `sigma_sol` in the function or the other way round.

include/GraphEdgeSolver.hpp
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+       @brief Solves \f$ (D M^{-1} D^T) u = g\f$, \f$ \sigma = M^{-1} D^T u \f$.
+
+       Dense block version
+
+       @param rhs \f$ g \f$ in the formula above
+       @returns sigma_sol \f$ \sigma \f$ in the formula above
+    */
+    DenseMatrix Mult(const DenseMatrix& input) const;
```
chakshinglee, line 127: also mismatch of parameter name in documentation

include/GraphEdgeSolver.hpp
gelever: Yes, I noticed that too.  I will probably rename everything to `u_sol`, `sigma_sol` as it is more descriptive.

chakshinglee:
By the way, for the v_e table, I guess it's up to us to decide what format to store it.

chakshinglee:
Oops, I accidentally closed this PR. 

Ok, what I wanted to say is, if we only store on each row, a pair of vertices of an edge (essentially no `indptr` and `data` for the e_v table), the size of data file can be reduced quite a bit. Probably still need to store number of edges/vertices in the beginning of the file.

gelever:
The graphdata can definitely be simplified.  linalgcpp can read an adjacency list: https://github.com/gelever/linalgcpp/blob/20311fbae3581a96b5f1e0e8b5b29a373e8aab71/include/parser.hpp#L203

chakshinglee:
Good, then we may choose that format. Yea, I was thinking to reduce the size of the fv graph, and realize the sample v_e is also not small. 

Seems not very natural to me to have such a function in linear algebra library though.

testcode/test_Graph.cpp
```diff
@@ -0,0 +1,82 @@
...
+ ***********************************************************************EHEADER*/
+
+/**
+   Test Graph object, distributed vs global
+
+   edge maps should be the same
+
+   vertex maps don't have to be the same,
+   but should assign each vertex a unique global id
+*/
```
chakshinglee, line 23: I dont quite understand the purpose of this test. Seems you want to check if the local to global map of edge id obtained from the two approaches match or not. But I think we dont need to construct local to global map if we start with a local (distributed) graph? So for the `Graph` constructor that take local graph as input, the local to global maps can be set to empty.

If I remember correctly we need this map (for vertex) because in some cases the Fiedler vector (store in text file) is a global vector, so if we want to setup the correct rhs, we will need this map.

I dont remember when did we add the edge map to `Graph`. I cannot recall any use case of it. I found that it is used in `GraphUpscale::GetEdgeVector`, which extract a subvector of a global edge vector. But I am worried about the correct usage of this function, because for shared edges, simply getting the subvector can be wrong in some cases (say, if it is used to distribute rhs).

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
-    int ptr;
-    for (unsigned int i = 0; i < nAggs; ++i)
+    for (int face = 0; face < num_faces; ++face)
     {
-        nlocal_coarse_dofs = vertex_target[i].Width();
-        GetTableRow(Agg_vertex, i, local_fine_dofs);
-        nlocal_fine_dofs = local_fine_dofs.Size();
-        mfem::DenseMatrix& target_i = vertex_target[i];
-        for (int j = 0; j < nlocal_fine_dofs; ++j)
+        std::vector<int> face_dofs = face_edge.GetIndices(face);
```
chakshinglee, line 176: I suggest we check here if size of `face_dofs` is 1, in which case we can send an empty matrix and move on (to the next face). Similarly for `CollectD` and `CollectM`. 

This can avoid some unnecessary communication when we have two big aggregates that connect to another one through only one edge.  I added this condition in the original code when I encountered some big graph (it was a YouTube graph) that has some hubs.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+
+    for (int i = 0; i < local_size; ++i)
+    {
+        assert(global_marker[local_indices[i]] >= 0);
+        row_map[i] = global_marker[local_indices[i]];
+    }
+
+    ClearMarker(global_marker, ext_indices);
+
+    return ext_mat.GetRow(row_map);
```
chakshinglee, line 994: (This should be pretty minor)

The original implementation is to first take transpose, extract columns, and then transpose the columns, which is copied from `ParElag`.

I remember at some point I was curious why not directly extract rows (what you are doing here), so I compare the two approaches. I found that the `ParElag` approach is actually faster, albeit it does extra transpose action twice. It was out of my expectation, and I think it's probably due to the storage format of DenseMatrix (data are concatenated columns) that makes copying columns (using `std::copy`) much faster. Did you observe similar behavior? 

`DenseMatrix::GetCol` in `linalgcpp` is using `GetSubMatrix` though, I think it may worth to use `std::copy` for columns, especially for continuous column indices.

src/GraphCoarsen.cpp
```diff
@@ -13,788 +13,1100 @@
...
+
+    for (int face = 0; face < num_faces; ++face)
     {
-        Agg_dof_i = new int[nAggs + 1];
-        Agg_dof_i[0] = 0;
+        int num_face_edges = face_edge.RowSize(face);
+
+        if (!sec_face.IsOwnedByMe(face))
+        {
+            edge_targets_[face].SetSize(num_face_edges, 0);
```
chakshinglee, line 291: Is it right that this line is not needed?

gelever: I think it was to make sure multiplying by a matrix would be well defined, even if the width is zero.  But this can be removed once `Mult` automatically resizes dense matrices.

barker29 (changes requested):
I am recommending that we abandon this (and that @chakshinglee stop taking time reviewing it). I am hopeful to see some small, incremental PRs that implement some of this work in pieces.

src/MinresBlockSolver.cpp
```diff
@@ -20,311 +20,175 @@
...
-    edgedof_d_td.MultTranspose(rhs.GetBlock(0), true_rhs_.GetBlock(0));
+    edge_true_edge_.MultAT(rhs.GetBlock(0), true_rhs_.GetBlock(0));
     true_rhs_.GetBlock(1) = rhs.GetBlock(1);
+    true_sol_ = 0.0;
 
-    if (!use_W_ && myid_ == 0)
+    if (!use_w_ && myid_ == 0)
     {
-        true_rhs_.GetBlock(1)(0) = 0.0;
+        true_rhs_[0] = 0.0;
```
chakshinglee, line 149: should here be `true_rhs_.GetBlock(1)[0]`?

In fact, I saw in the multilevel code this value does not affect the solution. I dont understand what is going on. But this may have something to do with "iterative mode"? In the original code, if minres' `iterative_mode` is false, this value will affect the solution, but when `iterative_mode` is true, similar behavior as in this new code. On the other hand, the 0-th entry of the final solution will equal to the 0-th entry of the initial guess.

gelever: I noticed this too when implementing ML branch and it should be fixed now (in the `feedback-dev` branch anyway).

src/MinresBlockSolver.cpp
```diff
@@ -20,311 +20,175 @@
...
+MinresBlockSolver::MinresBlockSolver(const MixedMatrix& mgl)
+    : MinresBlockSolver(mgl, {})
 {
-    Init(M, D, W);
 }
 
-void MinresBlockSolver::Init(mfem::HypreParMatrix* M, mfem::HypreParMatrix* D,
-                             mfem::HypreParMatrix* W)
+MinresBlockSolver::MinresBlockSolver(const MixedMatrix& mgl, const std::vector<int>& elim_dofs)
+    : MGLSolver(mgl), M_(mgl.GlobalM()), D_(mgl.GlobalD()), W_(mgl.GlobalW()),
```
chakshinglee, line 33: I think we need to do elimination on `M_` as well?

gelever: For our examples, it doesn't seem to make a difference but yes it should be eliminated also.

testcode/test_Graph.cpp
gelever: You are correct, `GetEdgeVector` does not produce the intended results in parallel.  I have now removed it since it is not actually used anywhere.

src/MinresBlockSolver.cpp
chakshinglee: I know in the existing examples `elim_dofs` is always empty (so you also dont need to eliminate the columns of D). I just find it inconsistent with the old code.

PR #24, opened 2018-06-19
Encapsulate upscaling parameters in object [feature/upscale-parameters]

barker29:
This allows the calling sequences of many constructors/functions to be much shorter and simpler.


examples/mlmc.cpp
```diff
@@ -273,19 +273,25 @@ int main(int argc, char* argv[])
 
     // Create Upscaler and Solve
     unique_ptr<FiniteVolumeMLMC> fvupscale;
+    UpscaleParameters param;
```
chakshinglee: How do you think about initializing `UpscaleParameters` before command line option parser, and then put the relevant member of `param` (e.g. `param.spect_tol`) directly in `OptionParser`? Something like
```
args.AddOption(&(param.spect_tol), "t", ...
```
This way we can save a few lines initializing each of the parameters in the beginning of `main`, as well as the lines below that assign values. And this comment is relevant for most of the examples/tests files.

barker29: I like this idea, and I've taken it one step further. Let me know what you think.

examples/graphupscale.cpp
```diff
@@ -77,8 +83,7 @@ int main(int argc, char* argv[])
     // vertex_edge and coarse factor
     {
         const auto upscale = GraphUpscale(comm, vertex_edge, coarse_factor,
-                                          spect_tol, max_evects, dual_target,
-                                          scaled_dual, energy_dual, hybridization);
+                                          param);
```
chakshinglee: very minor issue, this line may join the previous line? Similar issue also for some other files, e.g. `FiniteVolumeMLMC.cpp`.

examples/mlmc.cpp
chakshinglee: The register function looks good to me (it saves another few lines in each example/test). My only concern would be that new users may be not aware of these command options when looking at the cpp files. Some documentation may help?

examples/finitevolume.cpp
```diff
@@ -73,21 +68,10 @@ int main(int argc, char* argv[])
...
-    bool scaled_dual = false;
-    args.AddOption(&scaled_dual, "-sd", "--scaled-dual", "-no-sd",
-                   "--no-scaled-dual", "Scale dual graph Laplacian by (inverse) edge weight.");
-    bool energy_dual = false;
-    args.AddOption(&energy_dual, "-ed", "--energy-dual", "-no-ed",
-                   "--no-energy-dual", "Use energy matrix in trace generation.");
     bool visualization = false;
     args.AddOption(&visualization, "-vis", "--visualization", "-no-vis",
                    "--no-visualization", "Enable visualization.");
+    upscale_param.RegisterInOptionsParser(args);
```
chakshinglee: I'm thinking adding a comment like
```
// upscaling options from command line, see UpscaleParameters::RegisterInOptionsParser for details
```

barker29:
Any objections to merging this?

chakshinglee (approved):
looks good to me

gelever (approved):

PR #27, opened 2018-08-20
multilevel data structure first steps [feature/multilevel-dev]

barker29:
All this PR does is make some `coarse_` and `fine_` member variables in the `Upscale` classes into `std::vector` so that in principle we could have several of them in a multilevel setting.

This passes all the tests and is very slightly closer to Stephan's multilevel implementation. I do not claim this is a great improvement.


chakshinglee (approved):
These changes look fine to me.

PR #28, opened 2018-08-21
modual vertex targets [general-eigenproblem]

barker29:
The idea is to construct vertex targets / solve eigenproblem in slightly more modular way.

This will allow for future generalization with non-diagonal M for multilevel implementation.

This PR builds on multilevel data structures PR #27 and that PR should be reviewed before this one - after it is merged this one's diff will look nicer.


chakshinglee (approved):
Like you said in the comment, `MixedBlockEigensystem` may need further refactoring. It will be clearer when the multilevel development keeps moving on.

PR #29, opened 2018-10-08
Add support for nondiagonal M [nondiagonal-eigenproblem]

barker29:
This PR replaces some of the M matrix assembly with `DenseMatrix` instead of `Vector` so that we no longer assume this is diagonal when we coarsen.

All the current two-level tests pass, but this does not include some of the things necessary for a multilevel implementation.


chakshinglee (commented):
I think the changes are good for the purpose of the branch. There are two things that I think we may just do them in this PR as well:
1. `HybridSolver` does not assume diagonal (Vector) element matrices. This is partially done. One of the assembly functions has been removed. But in `Init`, the `fine_level` parameter is still used to optimize the setup. Since the most expensive setup is in the assembly (inversion of local matrices), which is now not optimized anyway, I think it does not make a big difference whether the rest of the setup is optimized. Without this optimization the code will look cleaner.

2. combine `ElementMBuilder` and `FineMBuilder` as both of them now contain element matrices of `DenseMatrix` type.

If you agree these should be done in this PR, I can try to make some commit for these two things. Otherwise I can do it in another PR.

By the way, for some reason something in travis is pending, do you know what is going on there?

barker29:
You can go ahead and do those things in this PR if you want.

I'm not sure why it shows one part of the check as pending, but if you click on "Details" next to it it seems like it is passing?

chakshinglee (approved):
@barker29 I have made the changes that we agreed, please take a look and see if you want any other modification.

PR #30, opened 2018-10-09
multilevel development [coarse-constant-representation]

barker29:
Some more features to support multilevel hierarchies.


src/MatrixUtilities.hpp
```diff
@@ -254,6 +254,8 @@ void Deflate(mfem::DenseMatrix& a, const mfem::Vector& v);
 */
 void orthogonalize_from_constant(mfem::Vector& vec);
 
+void orthogonalize_from_vector(mfem::Vector& vec, const mfem::Vector& wrt);
```
barker29, line 253: I should add a line of documentation.

src/MatrixUtilities.hpp
```diff
@@ -254,6 +254,8 @@ void Deflate(mfem::DenseMatrix& a, const mfem::Vector& v);
 */
 void orthogonalize_from_constant(mfem::Vector& vec);
```
barker29: Is this still necessary?

barker29 (commented):
I also need to add a test for `lineargraphthree`

src/MatrixUtilities.hpp
chakshinglee: I guess no. `orthogonalize_from_vector` should be general enough to do what `orthogonalize_from_constant` can do.

src/Upscale.hpp
```diff
@@ -107,39 +107,25 @@ class Upscale : public mfem::Operator
     virtual void OrthogonalizeCoarse(mfem::Vector& vect) const;
     virtual void OrthogonalizeCoarse(mfem::BlockVector& vect) const;
 
-    /// Create a coarse vertex space vector
-    virtual mfem::Vector GetCoarseVector() const;
+    virtual void OrthogonalizeLevel(int level, mfem::Vector& vect) const;
```
chakshinglee: Like `GetMatrix`, `GetVector`, etc. in this class, can we collapse each of `Orthogonalize`, `Restrict`, `Interpolate`, `BlockOffsets` and possibly `Solve` to only one that takes `level` as an input argument.

chakshinglee (approved):

PR #31, opened 2018-10-23
Local saddle point solver and assembling of coarse operators for general M

chakshinglee:
This PR is another step towards multilevel coarsening. Two main changes are:

1. `LocalGraphEdgeSolver` no longer assumes diagonal M, and the orthogonalization is also performed against user-provided representation of "constant", not hard-coded constant 1.

2. In `GraphCoarsen`, assembling of coarse (element) M and D is modified to accommodate non-diagonal finer-level M.

After these modifications, multilevel coarsening with "max_evects = 1" seems to work.

For example, the solution of `lineargraphthree` now looks reasonable:
```
./lineargraphthree --size 64 --partitions 32 --coarse-factor 2
```

I also tested the `finitevolume` and `generalgraph` examples with 3 or more levels, and seems getting reasonable approximation (need to set `-m  1` in command line). 

If this is merged, I think next step is generalization to higher order coarsening, which requires appropriate aggregation of dofs.

### DIFF
--- a/src/GraphCoarsen.cpp
+++ b/src/GraphCoarsen.cpp
@@ -176,6 +176,12 @@ void GraphCoarsen::NormalizeTraces(std::vector<mfem::DenseMatrix>& edge_traces,
         edge_traces_f.GetColumnReference(0, PV_trace);
         double oneDpv = Dtransfer.InnerProduct(PV_trace, localconstant);
 
+        if (fabs(oneDpv) < 1e-10)
+        {
+            std::cout << "Warning: oneDpv is closed to zero, oneDpv = "
+                      << oneDpv << ", this may be due to bad PV traces!\n";
+        }
+
         if (oneDpv < 0)
         {
             sign_flip = true;
@@ -489,6 +495,7 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
                         if (loc_map.first != other_j && loc_map.first != j)
                         {
                             other_j = loc_map.first;
+                            // TODO: avoid repeated extraction in high order coarsening
                             auto tmp = ExtractRowAndColumns(M_proc_, facefdofs[j],
                                                             facefdofs[other_j], colMapper_);
                             Mbb.Swap(tmp);
--- a/src/GraphCoarsen.cpp
+++ b/src/GraphCoarsen.cpp
@@ -367,7 +367,7 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
     mfem::Vector local_rhs_trace0, local_rhs_trace1, local_rhs_bubble, local_sol, trace;
     mfem::Array<int> local_verts, local_fine_dofs, faces;
     mfem::Array<int> facecdofs, local_facecdofs;
-    mfem::Vector one;
+    mfem::Vector one, first_vert_target;
     mfem::SparseMatrix Mbb;
     for (unsigned int i = 0; i < nAggs; i++)
     {
@@ -388,7 +388,6 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
         local_rhs_trace1.SetSize(nlocal_verts);
 
         mfem::DenseMatrix& vertex_target_i(vertex_target[i]);
-        double scale = vertex_target_i(0, 0);
 
         // ---
         // solving bubble functions (vertex_target -> bubbles)
@@ -454,8 +453,9 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
                 // compute and store local coarse D
                 if (k == 0)
                 {
+                    vertex_target_i.GetColumnReference(0, first_vert_target);
                     CoarseD_->Set(bubble_counter + i, row,
-                                  local_rhs_trace1.Sum() * -1.*scale);
+                                  (local_rhs_trace1 * first_vert_target) * -1.);
                 }
 
                 // instead of doing local_rhs *= -1, we store -trace later
--- a/src/GraphCoarsen.cpp
+++ b/src/GraphCoarsen.cpp
@@ -178,7 +178,7 @@ void GraphCoarsen::NormalizeTraces(std::vector<mfem::DenseMatrix>& edge_traces,
 
         if (fabs(oneDpv) < 1e-10)
         {
-            std::cout << "Warning: oneDpv is closed to zero, oneDpv = "
+            std::cerr << "Warning: oneDpv is closed to zero, oneDpv = "
                       << oneDpv << ", this may be due to bad PV traces!\n";
         }
 
--- a/src/GraphCoarsen.cpp
+++ b/src/GraphCoarsen.cpp
@@ -365,9 +365,10 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
     mfem::DenseMatrix traces_extensions, bubbles, B_potentials, F_potentials;
     mfem::Vector ref_vec1, ref_vec2;
     mfem::Vector local_rhs_trace0, local_rhs_trace1, local_rhs_bubble, local_sol, trace;
-    mfem::Array<int> local_verts, facefdofs, local_fine_dofs, faces;
+    mfem::Array<int> local_verts, local_fine_dofs, faces;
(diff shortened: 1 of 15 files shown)


testcode/coarse_assembling.cpp
```diff
@@ -0,0 +1,196 @@
...
+ * availability, see https://www.github.com/llnl/smoothG.
+ *
+ * smoothG is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License (as published by the Free
+ * Software Foundation) version 2.1 dated February 1999.
+ *
+ ***********************************************************************EHEADER*/
+
+/**
+   @file finitevolume.cpp
```
barker29: This comment block should be changed to say what this test actually does (ie, test that RAP is equal to the actually assembled D matrix).

testcode/CMakeLists.txt
```diff
@@ -78,6 +81,9 @@ add_test(rescaling rescaling)
 add_test(parrescaling mpirun -np 2 ./rescaling)
 add_valgrind_test(vrescaling rescaling)
 
+add_test(coarse_assembling coarse_assembling -m 1 --perm ${SPE10_PERM})
```
barker29, line 84: Thanks for adding tests! (we should also have some kind of `lineargraphthree` test but that will be my responsibility in the parent branch).

src/MatrixUtilities.hpp
```diff
@@ -341,40 +343,65 @@ class LocalGraphEdgeSolver
...
+                     \f$ \sigma = M^{-1} (D^T u + g) \f$.
 
-       @param rhs \f$ g \f$ in the formula above
+       @param rhs0 \f$ g \f$ in the formula above
+       @param rhs \f$ f \f$ in the formula above
        @param sol_sigma \f$ \sigma \f$ in the formula above
+       @param sol_u \f$ u \f$ in the formula above
     */
-    void Mult(const mfem::Vector& rhs, mfem::Vector& sol_sigma);
+    void Mult(const mfem::Vector& rhs0, const mfem::Vector& rhs1,
```
barker29:
```suggestion
    void Mult(const mfem::Vector& rhs_sigma, const mfem::Vector& rhs_u,
```

src/MatrixUtilities.hpp
```diff
@@ -341,40 +343,65 @@ class LocalGraphEdgeSolver
...
+    */
+    void Mult(const mfem::Vector& rhs, mfem::Vector& sol_sigma) const;
 
     /**
-       @brief Solves \f$ (D M^{-1} D^T) u = g\f$, \f$ \sigma = M^{-1} D^T u \f$.
+       @brief Solves \f$ (D M^{-1} D^T) u = f - D M^{-1} g \f$,
+                     \f$ \sigma = M^{-1} (D^T u + g) \f$.
 
-       @param rhs \f$ g \f$ in the formula above
+       @param rhs0 \f$ g \f$ in the formula above
```
barker29:
```suggestion
       @param rhs_sigma \f$ g \f$ in the formula above
```

src/MatrixUtilities.hpp
```diff
@@ -341,40 +343,65 @@ class LocalGraphEdgeSolver
...
+       @param rhs \f$ f \f$ in the formula above
```
barker29:
```suggestion
       @param rhs_u \f$ f \f$ in the formula above
```

src/MatrixUtilities.hpp
```diff
@@ -341,40 +343,65 @@ class LocalGraphEdgeSolver
...
-    void Mult(const mfem::Vector& rhs, mfem::Vector& sol_sigma);
+    void Mult(const mfem::Vector& rhs0, const mfem::Vector& rhs1,
+              mfem::Vector& sol_sigma, mfem::Vector& sol_u) const;
 
     /**
-       @brief Solves \f$ (D M^{-1} D^T) u = g\f$, \f$ \sigma = M^{-1} D^T u \f$.
+       @brief Solves \f$ (D M^{-1} D^T) u = f\f$, \f$ \sigma = M^{-1} D^T u \f$.
 
-       @param rhs \f$ g \f$ in the formula above
+       @param rhs1 \f$ f \f$ in the formula above
```
barker29:
```suggestion
       @param rhs_u \f$ f \f$ in the formula above
```

src/MatrixUtilities.hpp
```diff
@@ -341,40 +343,65 @@ class LocalGraphEdgeSolver
...
        @param sol_sigma \f$ \sigma \f$ in the formula above
        @param sol_u \f$ u \f$ in the formula above
     */
-    void Mult(const mfem::Vector& rhs, mfem::Vector& sol_sigma, mfem::Vector& sol_u);
+    void Mult(const mfem::Vector& rhs1,
```
barker29:
```suggestion
    void Mult(const mfem::Vector& rhs_u,
```

src/MatrixUtilities.cpp
```diff
@@ -964,85 +964,173 @@ void GenerateOffsets(MPI_Comm comm, int local_size, mfem::Array<HYPRE_Int>& offs
...
+        solver_->Mult(*rhs_, *sol_);
+        sol_sigma = sol_->GetBlock(0);
+    }
 
-    // Set rhs(0) back to its original vale (rhs is const Vector)
+    // Set rhs_block1(0) back to its original vale (rhs is const BlockVector)
     rhs_copy(0) = rhs_0;
+}
+
+void LocalGraphEdgeSolver::Mult(const mfem::Vector& rhs0, const mfem::Vector& rhs1,
```
barker29: Could change `rhs0` -> `rhs_sigma` etc. here, but it's less important than in the .hpp file.

src/GraphCoarsenBuilder.cpp
```diff
@@ -130,14 +130,14 @@ void ElementMBuilder::SetTraceBubbleBlock(int l, double value)
 
 void ElementMBuilder::AddTraceTraceBlockDiag(double value)
 {
-    M_el_[agg_index_](dof_loc_, dof_loc_) = value;
+    M_el_[agg_index_](dof_loc_, dof_loc_) += value;
```
barker29, line 133: How was this working before? (nice catch!)

src/GraphCoarsen.cpp
```diff
@@ -176,6 +176,12 @@ void GraphCoarsen::NormalizeTraces(std::vector<mfem::DenseMatrix>& edge_traces,
         edge_traces_f.GetColumnReference(0, PV_trace);
         double oneDpv = Dtransfer.InnerProduct(PV_trace, localconstant);
 
+        if (fabs(oneDpv) < 1e-10)
+        {
+            std::cout << "Warning: oneDpv is closed to zero, oneDpv = "
```
barker29:
```suggestion
            std::cerr << "Warning: oneDpv is closed to zero, oneDpv = "
```

src/GraphCoarsen.cpp
```diff
@@ -417,42 +423,55 @@ void GraphCoarsen::BuildPEdges(std::vector<mfem::DenseMatrix>& edge_traces,
         traces_extensions.SetSize(nlocal_fine_dofs, nlocal_traces);
         F_potentials.SetSize(nlocal_verts, nlocal_traces);
         local_facecdofs.SetSize(nlocal_traces);
+        local_rhs_trace0.SetSize(nlocal_fine_dofs);
+
+        std::vector<mfem::Array<int> > facefdofs(faces.Size());
+        std::vector<std::pair<int, int> > agg_trace_map(nlocal_traces);
```
barker29, line 429: This (and the changes to this routine more generally) is a very nice way to make this work.

barker29 (commented):
I'm very happy with the work you've done here.

On my local machine, the test `sampler` fails with a difference in max-p-error in the third decimal place. I'm not too worried about this - we'll rebaseline the test if we need to. Do you also see this difference? (I guess this test is not in Travis?)

chakshinglee:
@barker29 Actually, the test on `sampler` also fails on my machine. I ignored it because I thought the difference is due to the random number generator again (e.g. for tests involving graph generator, metis, results on mac OS are different). 

I do see a test named `sampler` (test #48) in the latest travis output, is this the test that you are referring to?

barker29:
I see it is passing in Travis. The different results locally are annoying, but it is a minor issue, and is a separate issue from this PR. We'll ignore it for now.

barker29 (approved):

PR #33, opened 2018-10-29
Combine Upscale classes 

chakshinglee:
Because the core part of all the *Upscale classes are essentially the same (the differences being how we construct the distributed graph and treatment of boundary conditions), this PR collapses them into Upscale. The benefit is to avoid changing all the *Upscale files whenever there is some update to Topology/Coarsener/Solver. 

Brief summary of changes:
1. Extract `ParGraph` class from `utilities.xpp` and rename it to simply `Graph`, which contains distributed vertex edge relation, edge weight, and edge to true edge relation. `Graph` can be constructed by distributing a global graph, or distributed graph. 
2. Use `Graph` as input parameter for `Upscale`, `GraphTopology`, `MixedMatrix`,  coarsener etc.
3. Move the construction of extended aggregates from `GraphTopology` to inside coarsener.
4. Remove `partitioning` in `Graph` because it should not be part of a graph. We had it before because we wanted to preserve the global partitioning when distributing the global graph, so that serial and parallel run will have the same upscaling error (one way to check if the parallel code is correct). Now partitioning is not preserved, parallel and serial run do not have the same results anymore, so some test criteria are rebaselined.  
5. Some renaming of variables (try to distinguish entities and dofs), and extract some steps into subroutines in the computation of targets.


examples/stest.py
```diff
@@ -581,7 +644,7 @@ def make_tests():
           "--kappa", "0.01",
           "--num-samples", "2"],
          {"fine-mean-l1": 0.54961180496539375,
-          "max-p-error": 0.39696741537314961}]
+          "max-p-error": 0.39869063097389679}]
```
chakshinglee, line 647: @barker29 In some PR you raised the concern on this mismatch. Does this update match with test result on your machine (up to `1e-4`)? While working on this PR, I realize that the test tolerance on travis is `1e-2` while the default tolerance specified in CMakeLists.txt is `1e-4`, I guess that's why the test is passing on travis but not on our machines.

examples/CMakeLists.txt
```diff
@@ -42,7 +42,7 @@ target_link_libraries(mlmc smoothg ${TPL_LIBRARIES})
 
 # add tests
 if (NOT DEFINED SMOOTHG_TEST_PROCS)
-    set(SMOOTHG_TEST_PROCS 4)
+    set(SMOOTHG_TEST_PROCS 2)
```
chakshinglee, line 45: @barker29 The test results on parallel graph now depend on number of processors, so I match this with the value set on travis.

barker29:
I haven't looked very closely, but my first concern is that the errors in many of the tests are much higher in this PR. Is this just because of the number of processors? I don't like an error of 0.8, to me this feels like the algorithm is somehow not working.

src/Upscale.hpp
```diff
@@ -37,6 +37,30 @@ namespace smoothg
 class Upscale : public mfem::Operator
 {
 public:
+    Upscale(const Graph& graph,
```
barker29: Can we get a line of documentation for each of these constructors? (also, it might be possible to combine a couple of them by using more default arguments?)

src/MixedMatrix.hpp
```diff
@@ -49,26 +49,7 @@ class MixedMatrix
...
-    MixedMatrix(const mfem::SparseMatrix& vertex_edge,
-                const mfem::Vector& weight,
-                const mfem::Vector& w_block,
-                const mfem::HypreParMatrix& edge_d_td,
-                DistributeWeight dist_weight = DistributeWeight::True);
-
-    MixedMatrix(const mfem::SparseMatrix& vertex_edge,
-                const std::vector<mfem::Vector>& local_weight,
-                const mfem::HypreParMatrix& edge_d_td);
+    MixedMatrix(const Graph& graph, const mfem::SparseMatrix& w_block = SparseIdentity(0));
```
barker29: This is a very nice simplification (in general your use of the `Graph` object is very nice and simplifies a lot!)

src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -167,6 +167,11 @@ class LocalMixedGraphSpectralTargets
         const mfem::SparseMatrix* W_local,
         const GraphTopology& graph_topology);
 
+    LocalMixedGraphSpectralTargets(
```
barker29, line 174: One line of documentation please (this one is actually a bit different from the other constructors, at least the calling sequence).

src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -182,26 +187,45 @@ class LocalMixedGraphSpectralTargets
                  std::vector<mfem::DenseMatrix>& local_vertex_targets,
                  const mfem::Vector& constant_rep);
 private:
+    enum DofType { vdof, edof }; // vertex-based and edge-based dofs
```
barker29: I think maybe `vdof` should be `VDOF` - that's the style I usually see for `enum`, but I guess we can make our own decision. What do you think?

src/GraphTopology.hpp
```diff
@@ -78,15 +92,14 @@ class GraphTopology
...
+                  const mfem::SparseMatrix& face_edge,
                   const mfem::SparseMatrix& Agg_vertex,
                   const mfem::SparseMatrix& Agg_edge,
-                  const mfem::HypreParMatrix& pAggExt_vertex,
-                  const mfem::HypreParMatrix& pAggExt_edge,
                   const mfem::SparseMatrix& Agg_face,
-                  const mfem::HypreParMatrix& edge_d_td,
-                  const mfem::HypreParMatrix& face_d_td,
-                  const mfem::HypreParMatrix& face_d_td_d);
+                  const mfem::HypreParMatrix& edge_trueedge,
```
barker29, line 100: These name changes are much better than before.

examples/stest.py
```diff
@@ -246,22 +246,43 @@ def make_tests():
           "--spect-tol", "1.0",
           "--max-evects", "1",
           "--coarse-components"],
-         {"finest-div-error": 0.37918423727222522,
-          "finest-p-error": 0.38013398274257243,
-          "finest-u-error": 0.38079825403520218,
+         {"finest-div-error": 0.79929158208615803,
+          "finest-p-error": 0.80196594708999625,
```
barker29, line 250: As I said in a previous comment, these errors seem really quite high. I want to understand why before I approve this PR.

examples/stest.py
barker29: Yes, this matches.

examples/stest.py
```diff
@@ -246,22 +246,43 @@ def make_tests():
...
+          "finest-u-error": 0.78297677149740286,
           "operator-complexity": 1.016509834901651}]
 
+    tests["parsamplegraph1-coeff-mac"] = \
```
barker29, line 254: These `-mac` tests are starting to get a bit out of hand. I'd like to have consistent tests that don't depend on platform, but that's probably a harder task and anyway would be a totally different PR.

src/Graph.cpp
```diff
@@ -0,0 +1,340 @@
+/*BHEADER**********************************************************************
+ *
+ * Copyright (c) 2017,  Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ * LLNL-CODE-XXXXXX. All Rights reserved.
```
barker29: This should have the correct release number (in general this should be done before anything is put on GitHub, even in a branch).

src/Graph.cpp
```diff
@@ -0,0 +1,340 @@
...
+ *
+ * This file is part of smoothG.  See file COPYRIGHT for details.
+ * For more information and source code availability see XXXXX.
```
barker29: replace XXXXX

src/Graph.hpp
```diff
@@ -0,0 +1,160 @@
+/*BHEADER**********************************************************************
+ *
+ * Copyright (c) 2017,  Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ * LLNL-CODE-XXXXXX. All Rights reserved.
```
barker29: release number.

src/Graph.hpp
```diff
@@ -0,0 +1,160 @@
...
+namespace smoothg
+{
+
+/**
+    @brief Manage topological information for the coarsening
+
+    Extract the local submatrix of the global vertex to edge relation table
+    Each vertex belongs to one and only one processor, while some edges are
+    shared by two processors, indicated by the edge to true edge
+    HypreParMatrix edge_e_te
```
barker29:
```suggestion
    HypreParMatrix edge_trueedge
```

src/Graph.hpp
```diff
@@ -0,0 +1,160 @@
...
+                    const mfem::SparseMatrix& vertex_edge_global,
+                    const mfem::Vector& edge_weight_global);
+
+    /**
+       @brief distribute a global serial graph into parallel local subgraphs.
+
+       Based on the partition numbers of vertices in partition_global,
+       distribute a global graph into local subgraphs in each processors, each
+       vertex is local to one and only one processor, while edges can be shared
+       between processors, which is encoded in edge_e_te_.
```
barker29:
```suggestion
       between processors, which is encoded in edge_trueedge_.
```

src/Graph.hpp
```diff
@@ -0,0 +1,160 @@
...
+
+       @param comm the communicator over which to distribute the graph
+       @param vertex_edge_global describes the entire global unsigned graph
+       @param vertex_edge_global edge weight of global graph, if not provided,
+    */
+    Graph(MPI_Comm comm,
+          const mfem::SparseMatrix& vertex_edge_global,
+          const mfem::Vector& edge_weight_global = mfem::Vector());
+
+    Graph(const mfem::SparseMatrix& vertex_edge_local,
```
barker29, line 85: A line of documentation please (it's fine to have thorough documentation for one constructor and short doc for the others, but they should all have something.)

src/Graph.hpp
```diff
@@ -0,0 +1,160 @@
...
+          const mfem::HypreParMatrix& edge_trueedge,
+          const mfem::Vector& edge_weight_local = mfem::Vector());
+
+    Graph(const mfem::SparseMatrix& vertex_edge_local,
```
barker29, line 96: Document.

src/Graph.cpp
```diff
@@ -0,0 +1,340 @@
...
+Graph::Graph(const mfem::SparseMatrix& vertex_edge_local,
+             const mfem::HypreParMatrix& edge_trueedge,
+             const mfem::Vector& edge_weight_local)
+    : vertex_edge_local_(vertex_edge_local)
+{
+    // temporary work-around (TODO: make a copy function for HypreParMatrix)
+    unique_ptr<mfem::HypreParMatrix> trueedge_edge(edge_trueedge.Transpose());
+    edge_trueedge_.reset(trueedge_edge->Transpose());
+
+    if (edge_weight_local)
```
barker29: `edge_weight_local` is a reference here, not a pointer, I'm not sure this will do what you expect. Maybe `if (edge_weight_local.Size() > 0)` ?

src/LocalMixedGraphSpectralTargets.cpp
```diff
@@ -596,10 +604,10 @@ mfem::Vector** LocalMixedGraphSpectralTargets::CollectConstant(
     const mfem::Vector& constant_vect)
 {
     // Gelever uses face_trueface rather than facedof_truedof (?) (I think one of us is just labeling it wrong)
```
barker29:
```suggestion
```

src/LocalMixedGraphSpectralTargets.cpp
```diff
@@ -359,65 +427,48 @@ void MixedBlockEigensystem::ComputeEdgeTraces(mfem::DenseMatrix& evects,
...
+    GetTableRow(ExtAgg_dof_offd, iAgg, dofs_offd);
+
+    int num_ext_loc_dofs_diag = dofs_diag.Size();
+    dofs.SetSize(num_ext_loc_dofs_diag + dofs_offd.Size());
+    std::copy_n(dofs_diag.GetData(), num_ext_loc_dofs_diag, dofs.GetData());
+    for (int i = 0; i < dofs_offd.Size(); i++)
+        dofs[i + num_ext_loc_dofs_diag] = dofs_offd[i] + num_ext_dofs_diag;
+}
+
 void LocalMixedGraphSpectralTargets::ComputeVertexTargets(
```
barker29, line 450: You've managed to both add functionality and make this simpler and easier to read, which is impressive.

src/GraphTopology.hpp
```diff
@@ -55,8 +56,21 @@ class GraphTopology
        @param partition partition vector for vertices
        @param edge_boundaryattr boundary attributes for edges with boundary conditions
     */
-    GraphTopology(mfem::SparseMatrix& vertex_edge,
-                  const mfem::HypreParMatrix& edge_d_td,
+    GraphTopology(const mfem::SparseMatrix& vertex_edge,
+                  const mfem::HypreParMatrix& edge_trueedge,
```
barker29, line 60: Modify documentation above to match name change.

barker29 (changes requested):
This is a lot of work and a big improvement.

The only thing I'm really worried about is the change in errors in all the tests - I want to understand what is causing this before I approve.

chakshinglee:
@barker29 I understand your concern on the huge difference in the upscaling errors. I noticed this too when I change them, and I was also surprised by the results. Here is some more info:

An observation that suggests the code is working properly:
One way to make the partition in parallel run to be same as serial run would be to use 12 processors because serial run partition the global graph into 12 partitions. I observe the same results in parallel and serial runs (up to 8 digits) if this partition is used. 

The partition when number of processors = 2 is bad (uneven):
One thing I just checked is that, the sizes of the distributed graphs are very different when 2 processes (in our tests) are used. One processor has 75 vertices, while the other gets 925. 

Right now, the number of partitions locally is set to the intended number of global partitions divided by number of processors. Based on the above split (79, 925), we can expect pretty poor partition if we divide each local graph into 6 partitions.

If I set the number of local partitions to be also proportional to relative local graph size (effectively consider "coarsening factor"). Then the errors get better (about 55% for different error measures). 

To make sure it is really just the partitioning that is causing the difference, I dont mind going back to our original partitioning method -- partition global graph, and distribute according to the global partition. This is not the main purpose of the PR anyway. I changed the partitioning method because I extracted `Graph` out, and and I think it doesn't make sense to have partition as a member of `Graph` objects. We can make this change in a separate PR. What do you think?

Another thing is, as I mentioned to you earlier, once we add the "redistribution" code, we can use the global partition to redistribute in parallel run, then we should expect the same upscale errors as before (since the partitioning should be the same). This option would be another PR also.

src/Upscale.hpp
chakshinglee: You are right, I can use more default arguments. In a newer version only one constructor is left

src/Graph.cpp
chakshinglee: Good catch, what you said should be the right way.

src/LocalMixedGraphSpectralTargets.hpp
chakshinglee: Sounds good to me, this will be updated.

barker29 (approved):
Thanks for all the changes, especially the nice documentation.

Regarding the error issue, I'm convinced it's okay. In a future PR, let's changethe default so the number of local partitions is  proportional to local graph size (ie "coarsening factor"), and have the tests based on that.

PR #35, opened 2018-11-28
GraphTopology refactoring

chakshinglee:
This PR is a step to make `Mixed_GL_Coarsener` an "intermediate" tool for coarsening. The goal is that we do not need to keep them (in a multilevel hierarchy). Only the mixed systems and interpolation matrices will be kept. This PR only deals with `GraphTopology`. Like coarsener, the relation tables between levels are needed only at the time of coarsening, we do not need to keep them.

Reorganize `GraphTopology`: 
1. Separate relation tables that describe fine graph, relation tables that describe coarse graph, and relation tables that describe the connection between fine and coarse graphs. Only the last group is left explicitly in `GraphTopology`. The fine graph is given from input and can be accessed through a raw pointer, and the coarse graph will be output from `GraphTopology::Coarsen`.
2. Construct `GraphTopology` inside `SpectralAMG_MGL_Coarsener` because it is part of the coarsening algorithm.
3. Unify `GraphTopology`constructors. The constructor that takes a coarse graph (GraphTopology) as input is now not needed, as the coarse graph is separated from `GraphTopology` (of previous level), so it is viewed just like another graph. The other "partial constructor" is also removed. It was only used in `lineargraph` as a test for using some manually constructed topology relations. Although the constructor is removed, this test is still doing the same thing. The `GraphTopology` is constructed from fine graph (no coarsening in this stage). The relation tables in `GraphTopology` are then copied from the manually constructed ones.

Minor: there are also some renaming/removal of members/functions in `Mixed_GL_Coarsener`.

### DIFF
--- a/src/Upscale.hpp
+++ b/src/Upscale.hpp
@@ -186,12 +186,12 @@ public:
 
     const mfem::SparseMatrix& GetPsigma(int level) const
     {
-        return coarsener_[level]->get_Psigma();
+        return coarsener_[level]->GetPsigma();
     }
 
     const mfem::SparseMatrix& GetPu(int level) const
     {
-        return coarsener_[level]->get_Pu();
+        return coarsener_[level]->GetPu();
     }
 
     /// Create Fine Level Solver
--- a/src/Upscale.hpp
+++ b/src/Upscale.hpp
@@ -208,7 +208,7 @@ public:
 
 protected:
 
-    void Init(const Graph& graph, const mfem::Array<int>& partitioning);
+    void Init(const mfem::Array<int>* partitioning);
 
     void MakeVectors(int level)
     {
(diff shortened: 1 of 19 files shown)


src/Upscale.hpp
```diff
@@ -208,7 +208,7 @@ class Upscale : public mfem::Operator
 
 protected:
 
-    void Init(const Graph& graph, const mfem::Array<int>& partitioning);
+    void Init(const mfem::Array<int>* partitioning);
```
barker29, line 216: Why is this a pointer now? Can't it still be a reference (minor question).

barker29 (approved):
This looks very nice, I can't even really find anything to complain about in my review!

src/Upscale.hpp
chakshinglee: The reason for pointer is I want to use `nullptr` as a check in https://github.com/LLNL/smoothG/blob/d3777bc9726b29e5130d7e4a96243c66b1d28af9/src/SpectralAMG_MGL_Coarsener.cpp#L42-L49

So if someone do not have a partitioning vector (for the first coarsening) provided to `Upscale`, `Upscale` will generate one based on `param_.coarse_factor`

constant reference would probably work also (in that case check if the `Size()` is 0). 

In general I have been sitting on the fence on whether to use `nullptr` or 0 size objects as default parameter to reduce number of functions, or maybe I should make them as separate functions. Do you have some preference on that?

PR #37, opened 2018-11-30
Multilevel sampling [working/combine multilevel refactor]

barker29:
This modifies the `PDESampler` object and the `sampler` and `mlmc` examples to be multilevel. It also adds a simple 3-level test for the `sampler` example.

### DIFF
--- a/examples/mlmc.cpp
+++ b/examples/mlmc.cpp
@@ -282,8 +282,8 @@ int main(int argc, char* argv[])
     unique_ptr<MultilevelSampler> sampler;
     if (std::string(sampler_type) == "simple")
     {
-        MFEM_ASSERT(num_levels == 2, "SimpleSampler only implemented 2-level here!");
-        sampler = make_unique<SimpleSampler>(vertex_edge.Height(), partitioning.Max() + 1);
+        std::vector<int> vertex_sizes = upscale.GetVertexSizes();
+        sampler = make_unique<SimpleSampler>(vertex_sizes);
     }
     else if (std::string(sampler_type) == "pde")
     {
--- a/examples/mlmc.cpp
+++ b/examples/mlmc.cpp
@@ -25,6 +25,9 @@
    ./mlmc --perm spe_perm.dat
 */
 
+// best multilevel command line so far (appears to create a reasonable result):
+// ./mlmc --sampler-type pde --num-samples 2 --max-levels 3 --hybridization --no-coarse-components --max-evects 1 --coarse-factor 8
+
 #include <fstream>
 #include <sstream>
 #include <mpi.h>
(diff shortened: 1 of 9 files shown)


examples/mlmc.cpp
```diff
@@ -275,18 +278,12 @@ int main(int argc, char* argv[])
...
-        std::cout << "fine graph vertices = " << num_fine_vertices << ", fine graph edges = "
-                  << num_fine_edges << ", coarse aggregates = " << num_aggs << std::endl;
-    }
-    unique_ptr<TwoLevelSampler> sampler;
+    const int num_levels = upscale_param.max_levels;
+    unique_ptr<MultilevelSampler> sampler;
     if (std::string(sampler_type) == "simple")
     {
-        sampler = make_unique<SimpleSampler>(num_fine_vertices, num_aggs);
+        MFEM_ASSERT(num_levels == 2, "SimpleSampler only implemented 2-level here!");
```
chakshinglee: Seems to me that you have made a multilevel version of `SimpleSampler` that takes a `vector` of sizes as input. The sizes should be available in `GraphTopology`, maybe need a function to put them into a `vector` though. So I think this can be in ML setting without too much effort.

chakshinglee: I was thinking a function in `Upscale` to collect and output those sizes.

examples/sampler.cpp
```diff
@@ -240,137 +255,156 @@ int main(int argc, char* argv[])
...
     if (save_statistics)
     {
-        SaveFigure(mean_upscaled, ufespace, "coarse_mean");
-        SaveFigure(mean_fine, ufespace, "fine_mean");
-        SaveFigure(m2_upscaled, ufespace, "coarse_variance");
-        SaveFigure(m2_fine, ufespace, "fine_variance");
+        SaveFigure(mean[1], ufespace, "coarse_mean");
+        SaveFigure(mean[0], ufespace, "fine_mean");
+        SaveFigure(m2[1], ufespace, "coarse_variance");
+        SaveFigure(m2[0], ufespace, "fine_variance");
```
chakshinglee: I am not sure the purpose of saving these figures, but we may want to save all the levels here?

examples/sampler.cpp
```diff
@@ -240,137 +255,156 @@ int main(int argc, char* argv[])
...
+                const double delta2 = (sol_upscaled(i) - mean[level](i));
+                m2[level](i) += delta * delta2;
+            }
+            p_error[level] = CompareError(comm, sol_upscaled, sol_fine);
+            if (myid == 0)
+            {
+                std::cout << "    p_error_level_" << level << ": " << p_error[level] << std::endl;
+            }
+
+            if (level == 1)
```
chakshinglee, line 338: same here, why only level 1 is saved?

chakshinglee (approved):
@barker29 the changes look good to me. There are several places seem can be put in multilevel setting without much extra work, I think it may worth doing that in this PR. It's up to your decision though.

PR #38, opened 2018-12-04
Introducing GraphSpace

chakshinglee:
First, this should be reviewed after the refactoring of `GraphTopology` is merged.

This PR adds `GraphSpace` to `smoothg`, which a collection of "entity to dof" relations. This can make things more organized in multilevel. Also, it makes the (hybridization) solver construction consistent in all levels. This is because the "entity to dof" relations are what actually needed for `HybridSolver`, we just have a special way to find them in the fine level so we did not ask for them. 

The following relations hold:
`GraphSpace` contains its associated `Graph` and a collection of "entity to dof" relations
`MixedMatrix` contains its associated `GraphSpace` and a collection of building blocks of the saddle point problem of the form [M  D^T \\\\ D  -W ].

Basically the `Mixed_GL_Coarsener` acts as a factory to produce a coarse `MixedMatrix` from a fine `MixedMatrix`, based on `UpscaleParameters` and partitioning array if provided. In the process, the interpolation matrices are constructed as well.

Another change is that I put the notion of "boundary" into `Graph`, which does not make much sense for graphs, but it helps to make the interface cleaner. Also, smoothG is able to do upscaling for mixed finite element problems (I think we can do it without much trouble, just need to write some interface), so "boundary" makes sense for those problems.

So far, this PR does NOT change anything as far as algorithm goes. That means multilevel still only works on one dof per entity. But with the introduction of `GraphSpace`, we are ready to go to higher order coarsening. In fact, I am working on that, but the change in this PR is probably large enough to make it an independent PR.


src/GraphSpace.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+   For brevity, vdof/edof refer to vertex/edge-based degree of freedom
+*/
+class GraphSpace
+{
+public:
+    /**
+       @brief Construct GraphSpace from Graph
+
+       Entity and dof are one-to-one.
```
barker29:
```suggestion
       Each entity has exactly one dof (as on the finest level).
```

src/GraphTopology.hpp
```diff
@@ -46,72 +46,44 @@ class GraphTopology
...
-                  const mfem::SparseMatrix* edge_boundaryattr = nullptr);
-
     /**
        @brief Build agglomerated topology relation tables of a given graph
 
        All of this data is local to a single processor
 
        @param graph graph oject containing vertex edge relation
-       @param partition partition vector for vertices
        @param edge_boundaryattr boundary attributes for edges with boundary conditions
```
barker29:
```suggestion
```

src/GraphTopology.hpp
```diff
@@ -144,28 +107,22 @@ class GraphTopology
...
     mfem::SparseMatrix Agg_vertex_;
     mfem::SparseMatrix face_Agg_;
-    mfem::SparseMatrix Agg_face_;
     mfem::SparseMatrix face_edge_;
     ///@}
 
-    /// "face" to boundary attribute table
-    mfem::SparseMatrix face_bdratt_;
+    /// pointer to coarse graph
+    const Graph* coarse_graph_;
```
barker29: Two things: (1) can we make this a `shared_ptr`? (2) does the pointer itself have to be public?

src/GraphTopology.cpp
```diff
@@ -424,28 +349,12 @@ void GraphTopology::Init(const mfem::Array<int>& partition,
...
-    for (int i = 0; i < num_levels - 2; i++)
-    {
-        graph_topologies.emplace_back(graph_topologies.back(), coarsening_factor);
-    }
+    auto face_trueface = BuildEntityToTrueEntity(*face_trueface_face_);
 
-    return graph_topologies;
+    auto coarse_graph = make_unique<Graph>(
+                            Agg_face, *face_trueface, mfem::Vector(), face_bdratt.get());
+    coarse_graph_ = coarse_graph.get();
```
barker29: I don't like this. You make a `unique_ptr`, then get a raw pointer, and you pretend to return a `unique_ptr` but it's not really unique because this class retains a (publicly accessible!) raw pointer to the same object.

barker29 (changes requested):
Other than the pointer issues in `GraphTopology`, this is a very nice simplification and I think it's ready to go.

src/GraphTopology.hpp
chakshinglee: (1) yes, `coarse_graph_ ` should be a `shared_ptr` here.
(2) the only place we need this public access is in the test `lineargraph`, where we manually build the coarse graph and set it in `GraphTopology`. So if `coarse_graph_` is not public, we should make a "Setter" for it. I will make a change for this alternative.

chakshinglee:
@barker29 `coarse_graph_` is now a `shared_ptr` and private member. What do you think about the current form? 

Actually, there are some other public members (e.g. Agg_vertex_, face_edge) in `GraphTopology`. The only place that need to modify their values is again in `lineargraph`. Now I tend to think that maybe we should change the way the test is setup. Currently we manually setup the members in `GraphTopology`, and put them into `GraphTopology` and use them for coarsening. This is some sort of verification that `GraphTopology` does produce what we want (the manually constructed ones). But Another way to verify this explicitly would be to construct `GraphTopology` based on the partition vector, and compare the automatically constructed ones with the manually constructed ones.

If we change `lineargraph` in the way stated above, those public members can be set as private and accessed through some "Getter" functions. What do you think?

src/GraphTopology.cpp
```diff
@@ -351,10 +350,9 @@ std::unique_ptr<Graph> GraphTopology::Coarsen(const mfem::Array<int>& partitioni
     // Construct "face to true face" table
     auto face_trueface = BuildEntityToTrueEntity(*face_trueface_face_);
 
-    auto coarse_graph = make_unique<Graph>(
-                            Agg_face, *face_trueface, mfem::Vector(), face_bdratt.get());
-    coarse_graph_ = coarse_graph.get();
-    return coarse_graph;
+    coarse_graph_ = make_unique<Graph>(Agg_face, *face_trueface, mfem::Vector(),
```
barker29: Shouldn't this be `make_shared`? I am getting confused by this.

barker29 (approved):
This looks better. Regarding your question about the `lineargraph` test, I think your proposed changes would be an improvement, but they don't have to go into this PR.

src/GraphTopology.cpp
chakshinglee: Good catch. My bad and I will correct it.

PR #39, opened 2018-12-12
Multilevel coarsening with more that one coarse dof per coarse entity

chakshinglee:
This PR pretty much completes the necessary components for generating "high order" coarse spaces recursively. 

The key change is to aggregate dofs instead of vertices/edges during coarsening step, which is not too difficult with the help of `GraphSpace`. 

One additional thing that we need is `AggregateDof`, which tells what are the finer level vdofs or evofs associated with some aggregate or face (coarse entity). They are needed when computing targets and solving extension/bubble problems. And it is a generalization of `GraphTopology`. I tried to keep the addition as minimal as possible, so just make `AggregateDof` as a `struct` storing all those relation tables to make the interface clean.

The rest of the core changes is to compute extended aggregate to dofs (instead of entity), and and use the aggregate to dof (instead of entity) for local matrix extraction.

Some supplementary additions are the sparse eigensolver for non-diagonal M (dense solver can be pretty slow for high order) and projection from the coarse vertex space to piecewise constant (on aggregate), and the inverse mapping. The latter two are used in `PDESampler` when projecting the coarse solution to a piecewise constant function and so do some (nonlinear) function evaluation and then project back to the coarse vertex space.

One additional tests for ml higher order space is ml-sampler4 (using `max_evects = 4`). And the test on `coarse_assembling` is set to higher order `max_evects  = 3" as well, verifying the coarse matrices are same as doing RAP on fine matrices.

Edited: my bad that in the above when I said `AggregateDof` I actually meant `DofAggregate`.


examples/stest.py
```diff
@@ -649,10 +649,21 @@ def make_tests():
...
         [["./sampler",
           "--num-samples", "1",
-          "--max-levels", "3"],
+          "--max-levels", "3",
+          "--spect-tol", "1.0",
+          "--max-evects", "1"],
          {"p-error-level-1": 0.20833920382939719,
           "p-error-level-2": 0.40118441747952621}]
 
+    tests["ml-sampler4"] = \
```
barker29, line 671: This test doesn't run because it's not added to `examples/CMakeLists.txt` (around line 125). (we should maybe automate this better so you only have to add the test in one place).

src/MatrixUtilities.hpp
```diff
@@ -79,6 +79,17 @@ mfem::SparseMatrix AAt(const mfem::SparseMatrix& A);
 
 std::unique_ptr<mfem::HypreParMatrix> AAt(const mfem::HypreParMatrix& A);
 
+/**
+    @brief Compute the product A * B between SparseMatrix and HypreParMatrix
+*/
+std::unique_ptr<mfem::HypreParMatrix> ParMult(const mfem::HypreParMatrix& A,
```
barker29, line 87: I'm having a hard time understanding this function. Am I right in understanding that the `SparseMatrix` object has different data on each processor, this routine just reinterprets that data as a parallel object? Why don't we just store it as a `HypreParMatrix` all the time?

src/MixedMatrix.hpp
```diff
@@ -69,7 +69,9 @@ class MixedMatrix
                 std::unique_ptr<MBuilder> mbuilder,
                 std::unique_ptr<mfem::SparseMatrix> D,
                 std::unique_ptr<mfem::SparseMatrix> W,
-                mfem::Vector constant_rep);
+                mfem::Vector constant_rep,
+                mfem::SparseMatrix Ppw1,
```
barker29: Can we document `Ppw1` and `Qpw1` here? (the names are not helpful unless you already know...)

src/Upscale.hpp
```diff
@@ -140,15 +140,16 @@ class Upscale : public mfem::Operator
         return GetMatrix(level).GetConstantRep();
     }
 
+    mfem::Vector PWConstProject(int level, const mfem::Vector& x) const;
```
barker29, line 146: A line of documentation please (you can just copy the comments from `MixedMatrix.hpp` lines 373-374).

barker29 (commented):
This looks very good!

In addition to my comments on a few lines of code, I have two other comments:
- I'd like to see a three-level spectral test with `generalgraph` or `finitevolume`.
- For me, the `pargraph-usegenerator` test is failing (looks like a minor difference in operator complexity)

src/MatrixUtilities.hpp
chakshinglee: You are right. This function will interpret a `SparseMatrix` as a "block-diagonal" `HypreParMatrix` and then call `mfem::ParMult`. This kind of multiplication is used in several places (e.g. `GraphTopology`, `LocalMixedGraphSpectralTargets`), so I just make the process (actually just a few lines) as a function. I will add more details in the documentation.

For the question whether we should just store the `SparseMatrix` as a `HypreParMatrix`, I think most of the time when this function is called, the `SparseMatrix` object is either temporary or it is more convenient to be in that format (e.g. `MixedMatrix::D_` for local extraction). So changing the storage format may result in more changes.

barker29: That makes sense. Your solution is fine with me.

chakshinglee:
@barker29 I was trying to find out what is causing the difference in operator complexity in `pargraph-usegenerator`. I think that is because when computing the local extension problem in `GraphCoarsen`, I used the first eigenvector as the local constant representation in this PR (instead of taking subvector of global constant representation). They should be more or less the same, but somehow there is some slight difference causing difference in NNZ. I just made a commit to change that back to the setting before the PR. Can you check again if this test passes now?

By the way, the default tolerance used in the tests is `1e-4`, while the tolerance in travis is `1e-2`, should we make them consistent?

barker29:
It still fails:
```
{
  "finest-div-error": 0.19793223114421926,
  "finest-p-error": 0.28945323154741931,
  "finest-u-error": 0.29617999465399153,
  "operator-complexity": 1.1897781022189777
}

  (1/1) [pargraph-usegenerator] FAILED.
```
Actually all the numbers are a little different, complexity is the most different, the python expects `1.1895581044189558`.

I'm not too worried about this difference, but I would like tests to pass (just changing the `stest.py` is fine if we understand that it's a slightly different algorithm).

I would like them to be consistent, but that may not be easy. To me `1e-2` is too forgiving, I view Travis as just a rough quick test and the "real" tests run on my local machine. Ideally we would have better tests, so that tests fail only when something actually changes, and not just when some tiny numerical precision issue propagates through. On the other hand changing everything to `1e-2` seems like it will allow real bugs to escape notice.

chakshinglee:
Ok, this is actually not what I expected. Because on my tux box the operator complexity after the last commit is indeed `1.1895581044189558`, that means the results on our machine are different too (using the same code). I agree that we can just change the number in `stest.py`, but now I don't know which number to use as reference. And I want to know what is causing our results to be different.

For Travis, I also think `1e-2` is too loose. Does the tolerance affect the test speed that much? I thought most of the test time is spent in actual computation (coarsening, solving).

chakshinglee:
@barker29 I have added some documentation regarding the "piecewise-constant projection". Please let me know if they make sense (mainly in `Upscale.hpp` and `MixedMatrix.hpp`). Also, a 3-level `finitevolume` test is added. 

The OC difference in `pargraph-usegenerator` is not resolved. Since the OC results on our machines are different due to some small entries (magnitude < `1e-14`), I want to check if we have the same OC on commit 7fcdcc7. In this version the small entries are just tiny but not exactly zero in my machine, so I guess our results have a better chance to agree. My result in that version is
```
  "finest-div-error": 0.19793206480307646,
  "finest-p-error": 0.28945297306669671,
  "finest-u-error": 0.29617987947091357,
  "operator-complexity": 1.1898581014189857
``` 
If you have similar result (difference < `1e-4`), I will just use this as reference.

barker29:
Unfortunately on that commit my OC is `1.1900580994190058` which is a bit over the line for our tolerance. I think this is just annoying, not important. Maybe we can put a `1e-3` tolerance on the tests?

barker29 (approved):
I think this is ready to go now. Thanks for all the hard work!

PR #41, opened 2019-01-14
Edge Projection operator

chakshinglee:
Add the projection operator for edge space. Add also debug checks (when debug mode is on) on whether the `Proj_sigma_` is really a projection, and if the commutative property holds.

For the first check, given a random coefficient vector in the coarse space `random_vec`, its fine level representation is `Psigma_ * random_vec`, so `Proj_sigma_ * Psigma_ * random_vec` should return the same random vector. In other words, `Proj_sigma_ * Psigma_` should be an identity operator.

The second check is strange forward. Note that using the notation in the paper, pi^u = `Pu_` * `Pu_`^T, pi^sigma = `Psigma_` * `Proj_sigma_`

### DIFF
--- a/src/Mixed_GL_Coarsener.cpp
+++ b/src/Mixed_GL_Coarsener.cpp
@@ -51,6 +51,13 @@ void Mixed_GL_Coarsener::Interpolate(const mfem::BlockVector& coarse_vect,
     Pu_.Mult(coarse_vect.GetBlock(1), fine_vect.GetBlock(1));
 }
 
+void Mixed_GL_Coarsener::Project(const mfem::BlockVector& fine_vect,
+                                 mfem::BlockVector& coarse_vect) const
+{
+    Proj_sigma_.Mult(fine_vect.GetBlock(0), coarse_vect.GetBlock(0));
+    Pu_.MultTranspose(fine_vect.GetBlock(1), coarse_vect.GetBlock(1));
+}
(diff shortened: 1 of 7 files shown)


src/Mixed_GL_Coarsener.cpp
```diff
@@ -63,5 +70,63 @@ void Mixed_GL_Coarsener::Interpolate(const mfem::Vector& coarse_vect,
     Pu_.Mult(coarse_vect, fine_vect);
 }
 
+void Mixed_GL_Coarsener::Project(const mfem::Vector& fine_vect,
+                                 mfem::Vector& coarse_vect) const
+{
+    Pu_.MultTranspose(fine_vect, coarse_vect);
+}
+
+void Mixed_GL_Coarsener::Debug_tests(const mfem::SparseMatrix& D) const
```
barker29, line 79: I really like these tests. One worry I have is that some crazy graph topology will break everything, but this gives us at least a chance to catch that in debug mode.

barker29 (approved):
Looks good!

src/Mixed_GL_Coarsener.cpp
chakshinglee: Yes, these tests and tolerance may need to be adjusted later. We may want to use relative difference, and only print warning not assert.

PR #42, opened 2019-01-15
Dirichlet pressure condition

chakshinglee:
This PR mainly adds an example involving Dirichlet pressure condition and ability to solve this kind of problem. Summary of changes:
1. added example/tests, see `finitevolume.cpp` and command line option `--lateral-pressure`. See also `SPE10Problem`, the boundary condition is determined by `ess_attr`. Note that `ess_attr` marks attributes with essential edge condition, the rest is natural boundary condition (in this problem it is exactly Dirichlet condition on pressure). The newly added tests are on error of quantity of interest.
2. Modified solvers. For `MinresBlockSolver` the elimination process is moved inside the solver. `HybridSolver` is also adjusted (quite substantially) to deal with the new BC type. Algorithmically, the main change is that the elimination process now is on "true multiplier" (on the parallel hybridized matrix). Lastly, the orthogonalization step is now determined and done inside solvers (so various orthogonalization in `Upscale` is removed).
3. This is not necessary, but I also made `SPE10Problem` and its parent class `DarcyProblem` to be like an interface between graph problem and finite volume problem. So `DarcyProblem` can build a weighted graph associated with an FV problem, as well as RHS, Cartesian partitioning, visualization. Outside the class we only see graph. This simplifies several examples, e.g. `finitevolume.cpp`, `mlmc.cpp`, `timestep.cpp`, `sampler.cpp`, because the construction of graph problem is moved to `DarcyProblem`.


examples/mlmc.cpp
```diff
@@ -154,128 +102,31 @@ int main(int argc, char* argv[])
...
 
     // Create Upscaler and Solve
-    Upscale upscale(*graph, upscale_param, &partitioning, &ess_attr);
+    Upscale upscale(graph, upscale_param, &partitioning, &ess_attr);
 
     upscale.PrintInfo();
     upscale.ShowSetupTime();
 
     mfem::BlockVector rhs_fine(upscale.GetBlockVector(0));
     rhs_fine.GetBlock(0) = 0.0;
```
barker29:
```suggestion
     rhs_fine.GetBlock(0) = spe10problem.GetEdgeRHS();
```

examples/pde.hpp
```diff
@@ -0,0 +1,873 @@
...
+            break;
+        }
+    }
+
+    if (no_flow_bc)
+    {
+        mfem::LinearForm q(u_fes_.get());
+        q.AddDomainIntegrator(new mfem::DomainLFIntegrator(*source_coeff_));
+        q.Assemble();
+        rhs_u_ = q;
```
barker29, line 846: I think it makes sense to set `rhs_sigma_ = 0.0;` here as well.

examples/pde.hpp
```diff
@@ -0,0 +1,873 @@
...
+ * availability, see https://www.github.com/llnl/smoothG.
+ *
+ * smoothG is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License (as published by the Free
+ * Software Foundation) version 2.1 dated February 1999.
+ *
+ ***********************************************************************EHEADER*/
+
+/**
+   @file spe10.hpp
```
barker29: `spe10` -> `pde` (and in the next line)

examples/pde.hpp
```diff
@@ -0,0 +1,873 @@
...
+    double dx[3];
+    mfem::Vector transip(dx, 3);
+
+    T.Transform(ip, transip);
+
+    return transip(1) < (spe10_scale_ * 44 * 5) ? -value_ : value_;
+}
+
+/**
+   @brief Darcy's flow problem discretized in finite volume (TPFA)
```
barker29, line 346: This is a very nice addition.

src/HybridSolver.cpp
```diff
@@ -121,71 +120,56 @@ void HybridSolver::Init(
...
-    mfem::SparseMatrix multiplier_edgedof(smoothg::Transpose(edgedof_multiplier) );
+    mfem::SparseMatrix mult_edof(smoothg::Transpose(edgedof_multiplier) );
 
-    mfem::Array<int> j_array(multiplier_edgedof.GetJ(), multiplier_edgedof.NumNonZeroElems());
-    j_array.Copy(j_multiplier_edgedof);
+    mfem::Array<int> j_array(mult_edof.GetJ(), mult_edof.NumNonZeroElems());
+    j_array.Copy(multiplier_to_edof_);
 
-    Agg_multiplier_.Clear();
+    Agg_multiplier_.Clear(); // FIXME(CSL 1/14/2019): this line is not needed?
```
barker29: Can we remove either the line or the comment?

src/HybridSolver.cpp
```diff
@@ -300,44 +283,35 @@ void HybridSolver::AssembleHybridSystem(
...
-        // Save element matrix [C 0][M B^T;B 0]^-1[C 0]^T
-        Hybrid_el_[iAgg] = tmpHybrid_el;
+        H_proc.AddSubMatrix(local_multiplier, local_multiplier, Hybrid_el_[iAgg]);
     }
+
+    return H_proc;
 }
 
+
 /// @todo nonzero BC, solve on true dof
```
barker29: I think this comment might be out of date (aren't you solving on true dof now?)

barker29 (approved):
You have again managed to add functionality while making the code base smaller. This looks very good.

src/HybridSolver.cpp
chakshinglee: This is a confusing comment (I will update shortly). Now `HybridSolver` works  for nonzero Dirichlet BC, but it still does not work for nonzero Neumann BC (essential edge unknown). In fact, I think even `MinresBlockSolver` doesn't work in the case of nonzero Neumann BC. 

For the second half, the hybridized system is on true dof, but `HybridSolver::Mult` is still assuming that input vectors (on the unknowns of the original problem) are not based on true dof (unlike `MinresBlockSolver`).

PR #43, opened 2019-01-28
Add PDESampler::SetSample() [feature/sampler-markov-chain]

barker29:
For Markov chain Monte Carlo, we need to provide an external white noise to `PDESampler` rather than having it create the white noise internally. This PR provides that.

It also contains a little bit of cleanup for the documentation and interface of `PDESampler`.


src/Sampler.cpp
```diff
@@ -127,17 +127,29 @@ PDESampler::~PDESampler()
...
+    for (int i = 0; i < num_aggs_[0]; ++i)
+    {
+        state(i) = normal_distribution_.Sample();
+    }
+
+    SetSample(state);
+}
+
+/// @todo cell_volume should be variable rather than constant
+void PDESampler::SetSample(mfem::Vector& state)
```
chakshinglee: I think input of this function can be `const mfem::Vector&`?

chakshinglee (approved):
This looks good to me, thanks!

PR #44, opened 2019-01-29
Add MLMC example [feature/mlmc-manager]

barker29:
Previously, we really only had multilevel sampling without any real multilevel Monte Carlo. The previously named `mlmc.cpp` is now `samplesolve.cpp`, and we do some very basic multilevel Monte Carlo in `qoi.cpp`.


src/MLMCManager.cpp
```diff
@@ -0,0 +1,450 @@
...
+}
+
+double MLMCManager::GetEstimate() const
+{
+    double out = 0.0;
+    for (const auto& v : mean_)
+    {
+        out += v;
+    }
+    return out;
```
chakshinglee: this function may be simplified by calling `std::accumulate`:
```
return std::accumulate(mean_.begin(), mean_.end(), 0.0);
```

src/MLMCManager.cpp
```diff
@@ -0,0 +1,450 @@
...
+    MFEM_ASSERT(out.Size() == vec1.Size(), "Sizes do not work!");
+    // ((mfem::Vector) out) = vec1;
+    for (int i = 0; i < out.Size(); ++i)
+    {
+        out[i] = vec1[i];
+    }
+    return out;
+}
+
+void MLMCManager::CoarseSample(bool verbose)
```
chakshinglee: This shares a lot of code with `MLMCManager::FineSample`. 

Maybe we can think of `MLMCManager` and `Upscale` as two different objects in that `Upscale` provides a "M-level" hierarchy, `MLMCManager` can do a "m-level" MLMC simulation for any "m" <= "M". This way, at the coarsest level "m-1" we will always just call this function (in the case of fine level MC, `num_levels_` = 0, so this function will pretty much do what `FineSample` does, except the rhs, which is commented below).

For any level "k" < "m-1", the CorrectionSample will be call.

src/MLMCManager.cpp
```diff
@@ -0,0 +1,450 @@
...
+{
+    const int level = num_levels_ - 1;
+    const int pressure_block = 1;
+
+    sampler_.NewSample();
+    auto coarse_coefficient = sampler_.GetCoefficient(level);
+    fvupscale_.RescaleCoefficient(level, coarse_coefficient);
+    mfem::BlockVector rhs_coarse = RestrictToLevel(fvupscale_, level, rhs_fine_);
```
chakshinglee: Whenever a new coefficient is sampled, `rhs_fine_` and `rhs_coarse` are not changed. So this restriction step can be performed once, and the rhs for all levels are stored for repeated use (both here and `CorrectionSample`.

src/MLMCManager.hpp
```diff
@@ -0,0 +1,247 @@
...
+*/
+class QuantityOfInterest
+{
+public:
+    virtual ~QuantityOfInterest() {}
+    /**
+       It might make sense to combine coefficient, pressure in some
+       kind of "state" object (that may even manage the solve), for now
+       we will just require both all the time even if you use just one.
+    */
```
chakshinglee, line 37: Seems `coefficient` is not used in any derived class of `QuantityOfInterest`, are you thinking there is potentially some other quantity of interest that needs coefficient to evaluate?

Or you are thinking `QuantityOfInterest` as a map that takes a sample of `coefficient` and return the value of a certain quantity of interest (like you said, manage the solve as well)? In that case we probably dont need pressure solution because that depends on `coefficient`.

src/Upscale.cpp
```diff
@@ -183,7 +183,7 @@ void Upscale::SolveAtLevel(int level, const mfem::Vector& x, mfem::Vector& y) co
 
 mfem::Vector Upscale::SolveAtLevel(int level, const mfem::Vector& x) const
 {
-    mfem::Vector coarse_vect = GetVector(1);
+    mfem::Vector coarse_vect = GetVector(level);
```
chakshinglee, line 186: Good catch! How did this function not break before? Maybe this is never called for `level !=1`.

src/MLMCManager.cpp
```diff
@@ -0,0 +1,450 @@
...
+    fvupscale_.SolveAtLevel(fine_level, rhs_fine_local, sol_fine);
+    double fineq = qoi_.Evaluate(fine_coefficient, sol_fine);
+    temp_cost += fvupscale_.GetSolveTime(fine_level);
+    // auto error_info = fvupscale_.ComputeErrors(sol_upscaled, sol_fine);
+
+    const double scale = 1.0 / ((double) sample_count_[fine_level] + 1.0);
+    varsum_[fine_level] += scale * ((double) sample_count_[fine_level]) *
+                           ((fineq - upscaledq) - mean_[fine_level]) * ((fineq - upscaledq) - mean_[fine_level]);
+    mean_[fine_level] += scale * ((fineq - upscaledq) - mean_[fine_level]);
+    cost_[fine_level] += scale * (temp_cost - cost_[fine_level]);
```
chakshinglee: this few lines (318-322) can be grouped as a function and be called in both `CoarseSample` and and `CorrectionSample`, with input parameters `level` and qoi (for here it is `fineq - upscaledq`, for `CoarseSample` it is QoI of that level).

src/MLMCManager.cpp
```diff
@@ -0,0 +1,450 @@
...
+        std::cout << "    combined cost: " << temp_cost << std::endl;
+        std::cout << "    uMLMC estimate: " << GetEstimate() << std::endl;
+    }
+
+    if (sample_count_[fine_level] < dump_number_)
+    {
+        // does it make sense to log some information, like the QoI, to go with the visualization?
+        // (conceivably you could even ask qoi_ to make a special picture)
+
+        // for more informative visualization (is it more sensible to do this here or in the python?)
```
chakshinglee: If you only want log scale for visualization in GLVis, there is an option in GLVis (key "L" if I remember correctly)

examples/qoi.cpp
```diff
@@ -0,0 +1,251 @@
...
+    rhs_fine.GetBlock(1) = 0.0;
+    // it may make sense to have more general, not hard-coded, right-hand side
+    int index = 0.5 * scale * scale + 0.1 * scale;
+    rhs_fine.GetBlock(1)(index) = 1.0;
+    index = 0.6 * scale * scale + 0.75 * scale;
+    rhs_fine.GetBlock(1)(index) = -1.0;
+
+    std::unique_ptr<MultilevelSampler> sampler;
+    const int seed = argseed + myid;
+    sampler = make_unique<PDESampler>(
```
chakshinglee: Minor: use of pointer is not needed here.

examples/qoi.cpp
```diff
@@ -0,0 +1,251 @@
...
+        }
+        for (int sample = 0; sample < shared_samples; ++sample)
+        {
+            if (myid == 0)
+                std::cout << "---\nShared sample " << sample << "\n---" << std::endl;
+
+            mlmc.CorrectionSample(0, verbose);
+        }
+    }
+    else
```
chakshinglee, line 190: I think this case covers `upscale_param.max_levels == 2` as well

examples/qoi.cpp
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+
+    for (int sample = 0; sample < choose_samples; ++sample)
+    {
+        if (myid == 0)
+            std::cout << "---\nChoose sample " << sample << "\n---" << std::endl;
+
+        mlmc.BestSample(verbose);
+    }
```
chakshinglee, line 194: It probably makes sense to wrap L189-243 into a function 
(say, `MLMCManager::Simulate(int num_levels, bool verbose)`, where `num_levels` is the number of levels of the MLMC simulation, which can be different from the number of levels of `Upscale` object that MLMCManager is constructed from)

examples/qoi.cpp
```diff
@@ -0,0 +1,251 @@
...
+    args.AddOption(&choose_samples, "--choose-samples", "--choose-samples",
+                   "Number of samples where MLMC will choose to optimize.");
+    int argseed = 1;
+    args.AddOption(&argseed, "--seed", "--seed", "Seed for random number generator.");
+    int dimension = 2;
+    args.AddOption(&dimension, "--dimension", "--dimension",
+                   "Spatial dimension of simulation.");
+    double cell_volume = 10000.0;
+    args.AddOption(&cell_volume, "--volume", "--volume",
+                   "Volume of typical cell (for PDE sampler).");
```
chakshinglee: Should this be obtained from `DarcyProblem::CellVolume` or `pmesh_`? It should match with `scale`.

examples/qoi.cpp
```diff
@@ -0,0 +1,251 @@
...
+        mfem::Mesh mesh(scale, scale, mfem::Element::QUADRILATERAL, 1,
+                        scale * 100.0, scale * 100.0);
+        pmesh = make_unique<mfem::ParMesh>(comm, mesh);
+    }
+
+    // Construct a graph from a finite volume problem defined on the mesh
+    mfem::Array<int> ess_attr(dimension == 3 ? 6 : 4);
+    DarcyProblem fvproblem(*pmesh, ess_attr);
+    Graph graph = fvproblem.GetFVGraph();
+    Upscale upscale(graph, upscale_param);
```
chakshinglee: if `ess_attr` is not provided, no elimination of edofs will be performed, and `Upscale` will treat it as a regular graph (no weird edges). I am not sure what is the final discrete problem that it represents. Similarly for `PDESampler`. 

Also, `HybridSolver` breaks because of the mismatch in graph boundary and `ess_attr_` (which is `nullptr` if not provided. For these two reasons I suggest to put `ess_attr` as one of the parameters (in that case you probably want to give an explicit value for `ess_attr`.

chakshinglee (commented):
This new feature looks so cool! 

Other than the individual comments, I am thinking maybe it makes sense to give some parameters to the `MLMCManager` to make it something analogous to linear solvers, one is maximum number of samples (it is something like `fine_samples` etc in `qoi.cpp`), the other one is tolerance for "approximated" mean square error (which ignores discretization error), and the simulation will stop when the error reaches tolerance. If tolerance is zero, then the simulation is only controlled by number of samples.

src/MLMCManager.hpp
barker29: I have a use case in mind for including coefficient, but you are right it is not implemented now. I can take it out but I think the abstract interface makes sense to include it.

chakshinglee: I see, let this stays as it is right now then.

barker29:
Thanks for all your comments, they were very helpful!

I think I've addressed all of them - please take another quick look and approve/decline.

examples/qoi.cpp
```diff
@@ -0,0 +1,205 @@
...
+                       graph, upscale_param, nullptr, &ess_attr);
+
+    mfem::Vector functional(rhs_fine.GetBlock(1));
+    functional = 0.0;
+    double dscale = 1.0 / problem_size;
+    for (int i = 0; i < problem_size; ++i)
+    {
+        functional(i) = dscale;
+    }
+    mfem::Vector dummy;
```
chakshinglee: this line can be removed

examples/qoi.cpp
```diff
@@ -0,0 +1,205 @@
...
+    rhs_fine.GetBlock(1)(index) = -1.0;
+
+    const int seed = argseed + myid;
+    PDESampler sampler(dimension, cell_volume, kappa, seed,
+                       graph, upscale_param, nullptr, &ess_attr);
+
+    mfem::Vector functional(rhs_fine.GetBlock(1));
+    functional = 0.0;
+    double dscale = 1.0 / problem_size;
+    for (int i = 0; i < problem_size; ++i)
```
chakshinglee: My understanding is that your are defining a functional to compute some sort of average pressure on the bottom boundary (as stated in the description). This definition is not computing that in 3D? But we can measure anything, so it doesn't really matter, it's up to you whether to change this or not.

chakshinglee (approved):
@barker29 I just had some minor comments. The rest look great to me, thanks!

examples/qoi.cpp
```diff
@@ -0,0 +1,205 @@
...
+    Graph graph = fvproblem.GetFVGraph();
+    Upscale upscale(graph, upscale_param, nullptr, &ess_attr);
+
+    upscale.PrintInfo();
+    upscale.ShowSetupTime();
+    upscale.MakeSolver(0);
+
+    mfem::BlockVector rhs_fine(upscale.GetBlockVector(0));
+    rhs_fine.GetBlock(0) = 0.0;
+    rhs_fine.GetBlock(1) = 0.0;
```
chakshinglee: these two lines can be combined to `rhs_fine = 0.0;`

PR #45, opened 2019-01-31
fix ordering of faces

chakshinglee:
While implementing the nonlinear solver, I realize that the multilevel (num_levels > 2) solver doesn't converge in parallel run (while serial run works fine). I found that the issue was that, when computing the traces on a shared face, we combine the traces, M, D, etc from each processor, these are local objects that are based on the local "edofs" ordering. However, there is no guarantee that the i-th edof from processor 0 on a shared face and the i-th edof from processor 1 correspond to the same "true edof". 

To fix this problem, the shared faces (edges of coarse graph) is reordered during the creation of coarse graph. In each processor, the ordering of the reordered (shared) faces should match with the corresponding true face ordering. Then since the ordering of coarse edofs is based on coarse faces, the ordering of edofs will also match automatically.

A debug check is added to detect if the edofs satisfy the above condition

So far the effect of this fix can only be seen in the nonlinear solver branch (which is not finished), but this may affect other development, so I propose a fix first.


barker29 (approved):
This is a nice fix.

PR #46, opened 2019-02-14
Separate Hierarchy from Upscale

chakshinglee:
This PR is to make the logic in the code more like the setting intended in the multilevel paper. We have a way to create a multilevel hierarchy (the newly added class `Hierarchy`), and the hierarchy can be used for: numerical upscaling, multilevel sampler, MLMC, multilevel solver.

`Hierarchy` contains the mixed system and solver in all levels, as well as the interpolation matrices.

`Hierarchy` is pretty much the previous `Upscale`, except that in each level, `Hierarchy::Solve` simply calls the solver in that level (which is the deleted `Upscale::SolveAtLevel`; on the other hand, `Upscale::Solve` will restrict `rhs` to coarse level, solve at that level, and then interpolate `sol` to fine level. In principle, the action in `Upscale::Solve` is not needed in other applications.

`Mixed_GL_Coarsener` and `SpectralAMG_MGL_coarsener` are removed, their role (wrapper for coarsening `MixedMatrix`) is taken over by `Hierarchy::Coarsen`, and the interpolation matrices in all levels are stored directly in `Hierarchy`.

There are some other minor simplifications.


src/Hierarchy.hpp
```diff
@@ -0,0 +1,189 @@
...
+ *
+ * smoothG is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License (as published by the Free
+ * Software Foundation) version 2.1 dated February 1999.
+ *
+ ***********************************************************************EHEADER*/
+
+/** @file
+
+    @brief Contains Upscale class
```
barker29: `Hierarchy` class

src/MinresBlockSolver.cpp
```diff
@@ -239,72 +183,52 @@ void MinresBlockSolver::Mult(const mfem::BlockVector& rhs,
...
-    true_rhs_.GetBlock(1) = rhs.GetBlock(1);
+    const auto& edof_trueedof = mixed_matrix_.GetGraphSpace().EDofToTrueEDof();
+    edof_trueedof.MultTranspose(rhs.GetBlock(0), rhs_.GetBlock(0));
+    rhs_.GetBlock(1) = rhs.GetBlock(1);
 
-    if (!W_is_nonzero_ && remove_one_dof_ && myid_ == 0)
-    {
-        true_rhs_.GetBlock(1)(0) = 0.0;
-    }
+    MinresBlockSolver::Mult(rhs_, sol_);
```
barker29, line 217: These are some nice simplifications.

src/Sampler.cpp
```diff
@@ -58,57 +58,38 @@ mfem::Vector& SimpleSampler::GetCoefficient(int level)
...
-    fvupscale_(fvupscale),
-    normal_distribution_(0.0, 1.0, seed),
-    num_aggs_(fvupscale->GetNumLevels()),
-    cell_volume_(cell_volume),
-    sampled_(false),
-    rhs_(fvupscale->GetNumLevels()),
-    coefficient_(fvupscale->GetNumLevels())
+PDESampler::PDESampler(int dimension, double cell_volume, double kappa, int seed,
+                       Hierarchy&& hierarchy)
+    : hierarchy_(std::move(hierarchy))
```
barker29, line 63: Can a `PDESampler` and a `Upscale` use the same `Hierarchy`? (they both `std::move` it into their own class). I'm not sure we will need to do this, but it seems like we might.

barker29 (approved):
In general this looks like a very nice PR. On my desktop I see failures for tests `qoi`, `qoi-one-level`, and `qoi-hb`. The differences are not small, either (in fact I think they should show up in Travis if it is doing the same thing. For example, I get
```
MLMC estimate: -0.0157326
```
when `stest.py` expects `-0.0075878`.

I suspect this is just the usual random number things, but these tests pass on master. We probably need generally a better framework for testing things with random properties, rather than trusting that setting the seed makes everything reproducible.

src/Sampler.cpp
chakshinglee: I also thought about this when working on this PR, I was thinking the hierarchy in `PDESampler` has a nonzero "W block" and constant coefficient, while the hierarchy for `Upscale` or `MLMCManager` typically has some heterogeneous coefficient and no "W block".

But I agree there might be a case that the two hierarchies are the same. In this case, there is still one reason to keep them separate. In the applications that we consider, the coefficient in `Upscale` will be updated during simulation, but we do not change the coefficient in `PDESampler`.

barker29: That makes sense. We can leave it as you have it.

chakshinglee:
@barker29 I think what you caught is from an actual bug. My tux box also showed what you observed. They were passed on Travis again because of the loose tolerance `1e-2`. 

After 52eb10a (it's about the sign of right hand side or solution, in this PR I removed several "` *=  -1.0`" lines), then those tests are passing now. 

Can you check the tests again? Thanks!

barker29:
Looks better now, thank you!

PR #48, opened 2019-03-21
TPFA for general meshes

chakshinglee:
Previously we construct transmissibility (edge weight) by using mfem bilinear form integrator, it is not working as expected for triangle and tetrahedron - I think the normal and face area for triangle and tetrahedron are wrong in `FiniteVolumeMassIntegrator`. The normal and face area were extracted using RT0 basis, and this is okay on Cartesian grids. For simplices, I tried to print them out and they are not what I expected. 

The new class `LocalTPFA` in this PR directly computes the transmissibility using the mesh and coefficient (which are what TPFA originally depends on).

### DIFF
--- a/examples/pde.hpp
+++ b/examples/pde.hpp
@@ -356,10 +356,14 @@ class LocalTPFA
     const mfem::ParMesh& mesh_;
     const mfem::SparseMatrix& vert_edge_;
 
+    mfem::DenseMatrix EvalKappa(int i);
+    mfem::Vector ComputeShapeCenter(const mfem::Element& el);
+    mfem::Vector ComputeLocalWeight(int i);
 public:
     LocalTPFA(const mfem::ParMesh& mesh, const mfem::SparseMatrix& vert_edge)
         : Q_(NULL), VQ_(NULL), mesh_(mesh), vert_edge_(vert_edge) { }
 
+    /// @param q inverse of permeability \f$ kappa^{-1} \f$ in Darcy's law
     LocalTPFA(const mfem::ParMesh& mesh, const mfem::SparseMatrix& vert_edge,
               mfem::Coefficient& q)
         : Q_(&q), VQ_(NULL), mesh_(mesh), vert_edge_(vert_edge) { }
@@ -368,7 +372,7 @@ public:
               mfem::VectorCoefficient& q)
         : Q_(NULL), VQ_(&q), mesh_(mesh), vert_edge_(vert_edge) { }
 
-    mfem::Vector ComputeLocalWeight(int i);
+    /// Compute local edge weights for the corresponding graph Laplacian
     std::vector<mfem::Vector> ComputeLocalWeights()
     {
         std::vector<mfem::Vector> local_weights(mesh_.GetNE());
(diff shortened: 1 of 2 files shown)


examples/pde.hpp
```diff
@@ -342,6 +342,141 @@ double HalfCoeffecient::Eval(mfem::ElementTransformation& T,
...
+    mfem::Coefficient* Q_;
+    mfem::VectorCoefficient* VQ_;
+    const mfem::ParMesh& mesh_;
+    const mfem::SparseMatrix& vert_edge_;
+
+    mfem::DenseMatrix EvalKappa(int i);
+    mfem::Vector ComputeShapeCenter(const mfem::Element& el);
+    mfem::Vector ComputeLocalWeight(int i);
+public:
+    LocalTPFA(const mfem::ParMesh& mesh, const mfem::SparseMatrix& vert_edge)
```
barker29: I think this (and the other constructors) can build `vert_edge` from the `mesh`, as is currently done in `DarcyProblem::BuildReservoirGraph`.

It doesn't matter whether this is done here or in `DarcyProblem` (in fact I'm not too sure what the difference between them should be), but it would make this interface simpler.

examples/pde.hpp
```diff
@@ -342,6 +342,141 @@ double HalfCoeffecient::Eval(mfem::ElementTransformation& T,
...
+    else
+    {
+        for (int d = 0; d < dim; ++d)
+            kappa(d, d) = 1.0;
+    }
+
+    return kappa;
+}
+
+mfem::Vector LocalTPFA::ComputeShapeCenter(const mfem::Element& el)
```
barker29, line 416: I think it will be simpler (and more general) to call `mfem::Mesh::GetElementCenter()`).

examples/pde.hpp
```diff
@@ -342,6 +342,141 @@ double HalfCoeffecient::Eval(mfem::ElementTransformation& T,
...
+    const mfem::Vector cell_center = ComputeShapeCenter(*(mesh_.GetElement(i)));
+
+    mfem::Array<int> edges;
+    GetTableRow(vert_edge_, i, edges);
+
+    mfem::ParMesh& non_const_mesh = const_cast<mfem::ParMesh&>(mesh_);
+    mfem::Vector normal(dim), c_vector(dim);
+    mfem::Vector local_weight(num_facets);
+
+    for (int e = 0; e < num_facets; ++e)
```
barker29: Is there a reason you're using `facet` here? This is terminology invented by someone for Parelag, I think `face` is better in this method.

barker29 (approved):
Looks good, only minor comments.

examples/pde.hpp
chakshinglee: Yes, I also thought of doing that (was debating with myself). With your opinion now I know which option to choose.

examples/pde.hpp
chakshinglee: Yes, this is a better way. I was not aware of this function, thanks for pointing it out. By the way, do we stick with `mfem v3.3` (as in the `INTALL.md`) or just the latest version? Since I couldn't find `Mesh::GetElementCenter()` in `v3.3.2` (the one that I am using)?

examples/pde.hpp
chakshinglee: My interpretation on `facet` is that a `facet` of a n-dimensional polytope is a (n-1)-dimensional object on part of boundary (as in [wiki](https://en.wikipedia.org/wiki/Face_(geometry)#Facet_or_(n-1)-face)). So it fits in both 2D and 3D: 2D is an edge, 3D is a face. 

face should be clear enough from the context though (I think the reader can understand it as an edge in 2D), so I will make the corresponding change.

examples/pde.hpp
barker29: That's a good point. I'm apparently using a newer MFEM (which works fine) but I'm not sure this particular thing is a good reason to upgrade the whole project. Either way is fine with me.

examples/pde.hpp
barker29: I learned something new today! I guess either way is fine.

examples/pde.hpp
chakshinglee: Is there some similar function for computing face/edge center as well? If not, we have to keep this extra function because face/edge center are needed as well.

barker29: I don't see one, so maybe we should just leave it.

PR #49, opened 2019-05-24
Fix/fine edge order

chakshinglee:
Previously we have fixed the edge ordering (all local orderings follow true ordering) when we construct a coarse graph in `GraphTopology`. Turns out when we construct graph from `mfem::ParMesh`, the ordering sometimes has problem too. So this PR is to do the same fix as in `GraphTopology`.

### DIFF
--- a/examples/stest.py
+++ b/examples/stest.py
@@ -709,9 +709,9 @@ def make_tests():
           "--shared-samples", "18",
           "--choose-samples", "0",
           "--seed", "1"],
-         {"coarse-variance":0.00016563124437545969,
-          "correction-variance":3.6669929731798261e-05,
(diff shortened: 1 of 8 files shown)


barker29:
On my machine the tests `qoi` and `qoi-hb` are failing; there are seemingly significant differences in the first decimal place of the reported values. On master these tests pass. Did the random seed somehow get changed?

barker29 (changes requested):
For now let's just rebaseline the tests (ie, change the expected results in `stest.py`) since we know what is going on, and then I will approve. I'm working on another PR that will have slightly better tests.

chakshinglee:
I just rebaselined the tests based on the value in my machine. Hopefully they match with yours.

barker29 (approved):

Issue #47, opened 2019-03-08
Travis tests are too forgiving


barker29:
Our default test tolerance in Travis is too loose - some real bugs can get through (see PR #46). Part of the reason is because of platform differences, especially in random numbers, so we may need to fix Issue #32 and Issue #34 before this one, but the tolerance in Travis is an issue all on its own.

barker29:
Fixed in #50

PR #50, opened 2019-05-28
Allow stricter testing in Travis CI [fix/testing]

barker29:
Allow stricter testing in Travis CI.

This addresses Issue #47

### DIFF
--- a/examples/stest.py
+++ b/examples/stest.py
@@ -36,7 +36,7 @@ memorycheck_command = "@MEMORYCHECK_COMMAND@"
 
 # Test paramaters
 num_procs = "@SMOOTHG_TEST_PROCS@"
-test_tol = float("@SMOOTHG_TEST_TOL@")
+default_test_tol = float("@SMOOTHG_TEST_TOL@")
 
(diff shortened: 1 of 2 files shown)


examples/stest.py
```diff
@@ -66,10 +66,19 @@ def run_test(command, expected={}, verbose=False):
...
 
-    for key, expected_val in expected.items():
+    for key, expected_valpair in expected.items():
         test_val = output[key]
 
-        if abs(float(expected_val) - float(test_val)) > test_tol:
+        try:
+            expected_val = expected_valpair[0]
+            test_tol = expected_valpair[1]
+        except TypeError:
```
chakshinglee, line 75: If I understand correctly, here you use exception as "`if`": if expected_valpair has size 2, then do the `try` part, otherwise do the other part. This works fine, but I am wondering if we can use a normal "`if`" (the exception is not really an "error").

chakshinglee:
I like the new change. I have always wanted to know what is the expected value when a test is failed. Previously I had to go to `stest.py` to find that out.

chakshinglee (approved):

PR #51, opened 2019-05-30
Improve performance of HybridSolver

chakshinglee:
Two improvements: 
1. add a way to automatically compute a good scaling vector based on SIAM J. Sci. Comput., 41(3), B425–B447. This works well when there is only one "edge dof" per edge.

2. when there is more that one "edge dof" per edge in the upscaled problem, the scaling vector in 1 most likely will generate bunch of zeros (because of the way we construct coarse space, "divergence" of some of them are zero). So in this case, project the problem to the space of PV vectors (where we can solve efficiently because of point 1), and supplement this with some smoothing (block Jacobi) on the original problem.

Here are some results with the new solver:

two example runs on `master`:
```
mpirun -n 4 ./finitevolume -ma -s 84 -t 1 -m 5 -hb
Level 1 Solve Time:         0.0503802
Level 1 Solve Iterations:   213
```
and 
```
mpirun -n 4 ./finitevolume -ma -d 3 -t 1 -m 8 -hb
Level 1 Solve Time:         165.585
Level 1 Solve Iterations:   3496 
```

The same example on `feature/better-hybrid-solver`:
```
mpirun -n 4 ./finitevolume -ma -s 84 -t 1 -m 5 -hb
Level 1 Solve Time:         0.00631809
Level 1 Solve Iterations:   23
```
and 
```
mpirun -n 4 ./finitevolume -ma -d 3 -t 1 -m 8 -hb
Level 1 Solve Time:         0.620253
Level 1 Solve Iterations:   24
```

### DIFF
--- a/src/HybridSolver.cpp
+++ b/src/HybridSolver.cpp
@@ -515,7 +515,7 @@ void HybridSolver::ComputeScaledHybridSystem(const mfem::HypreParMatrix& H)
         mfem::HypreSmoother prec_scale(const_cast<mfem::HypreParMatrix&>(H));
 
         mfem::Vector zeros(H.Height());
-        zeros = 1e-8;
+        zeros = 0.0;
         diagonal_scaling_.SetSize(H.Height());
         diagonal_scaling_ = 1.0;
 
--- a/src/HybridSolver.cpp
+++ b/src/HybridSolver.cpp
@@ -35,8 +35,7 @@ HybridSolver::HybridSolver(const MixedMatrix& mgL,
                            const SAAMGeParam* saamge_param)
     :
     MixedLaplacianSolver(mgL.GetComm(), mgL.BlockOffsets(), mgL.CheckW()),
-    D_(mgL.GetD()),
-    W_(mgL.GetW()),
+    mgL_(mgL),
     rescale_iter_(rescale_iter),
     saamge_param_(saamge_param)
 {
@@ -50,9 +49,6 @@ HybridSolver::HybridSolver(const MixedMatrix& mgL,
     }
 
     const GraphSpace& graph_space = mgL.GetGraphSpace();
-
-    Agg_vertexdof_.MakeRef(graph_space.VertexToVDof());
-    Agg_edgedof_.MakeRef(graph_space.VertexToEDof());
 
     Init(graph_space.EdgeToEDof(), mbuilder->GetElementMatrices(),
          graph_space.EDofToTrueEDof(), graph_space.EDofToBdrAtt());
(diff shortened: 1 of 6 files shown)


src/HybridSolver.cpp
```diff
@@ -463,22 +510,28 @@ void HybridSolver::RecoverOriginalSolution(const mfem::Vector& HybridSol,
...
-
-    auto Scale = VectorToMatrix(diagonal_scaling_);
-    mfem::HypreParMatrix pScale(comm_, H.N(), H.GetColStarts(), &Scale);
-    H_.reset(smoothg::Mult(pScale, H, pScale));
+    if (rescale_iter_ > 0)
+    {
+        mfem::HypreSmoother prec_scale(const_cast<mfem::HypreParMatrix&>(H));
+
+        mfem::Vector zeros(H.Height());
+        zeros = 1e-8;
```
barker29: This seems like a strange choice for right-hand side. I guess the idea is to approximate a "near nullspace", so you want to solve something close to `Ax = 0`? But this is actually equivalent to solving `Ax = 1` I think. (I realize this choice does not belong to this PR, so it doesn't change my review.)

src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -117,6 +120,8 @@ class UpscaleParameters
                        "--no-coarse-components", "Store trace, bubble components of coarse M.");
         args.AddOption(&coarse_factor, "--coarse-factor", "--coarse-factor",
                        "Coarsening factor for metis agglomeration.");
+        args.AddOption(&rescale_iter, "--rescale-iter", "--rescale-iter",
```
barker29, line 123: What does this actually do? Is the hybridization solver more effective if this number is larger (at the cost of higher setup cost?)

barker29 (approved):
This is very nice with very impressive results!

You may want to add a citation of your paper in `README.md`, or maybe in the comments of `HybridSolver.hpp` if you want.

Also, I am starting to wonder if the hybridization solver should be the default all the time, maybe we should begin moving away from or deprecating the minres solver.

src/HybridSolver.cpp
chakshinglee: Yes, I agree that the current setup is equivalent to solving `Ax = 1` (if solve exactly). I now change the rhs to exact 0.

src/LocalMixedGraphSpectralTargets.hpp
chakshinglee: This number determines the way we compute the "rescaling vector" for the hybridized system. Like you said, we want a vector in the "near nullspace". Currently there are two ways to find it in the code. The first way is to approximate it by applying a few smoothing steps to `Ax = 0` with a nonzero initial guess (this was in the code before this PR). The second way to set it as some special vector based on Prop. 4.3 in SISC, 41(3), B425–B447 (newly added to this PR). 

When `rescale_iter` is negative, the rescaling vector is computed using the second way. This is now default because it seems to work pretty well, and it is very cheap to obtain (matrix-vector multiplication).

When `rescale_iter` is positive, the rescaling vector is computed using the first way, and `rescale_iter` indicates the number of smoothing steps.

When `rescale_iter` is 0, the hybridized system will not be rescaled.


To answer your second question, yes in general it is true that the solver is more effective when `rescale_iter` is larger.

PR #54, opened 2019-08-02
borrow dependencies from MFEM [simplify-cmake]

barker29:
Instead of specifying our own locations/libraries for HYPRE, Suitesparse, etc., this PR would just look up what dependencies MFEM uses and follow that.

This PR may not be a good idea - let me know what you think.

Advantages: 
- you can use MFEM with some additional packages (libunwind, gzstream, etc.) without having to add those to the build system of smoothG
- packages are guaranteed consistent between MFEM and smoothG

Disadvantages:
- we were actually trying to move *away* from an MFEM dependency, this PR ties us closer to them
- the configuration process (especially the new `parsemk.py`) may be less flexible, might break in unexpected ways

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ generated-docs/
 *.md~
 *.swp
 
+out.cmake
 GPATH
 GRTAGS
 GTAGS
--- a/.travis.yml
+++ b/.travis.yml
@@ -160,9 +160,6 @@ script:
    - cd build;
    - |
        CC=mpicc CXX=mpic++ cmake -DMFEM_DIR="/home/travis/build/LLNL/mfem-install" \
-       -DHYPRE_DIR="/home/travis/build/LLNL/hypre-install" \
-       -DSuiteSparse_DIR="/home/travis/build/LLNL/SuiteSparse-install" \
-       -DMETIS_DIR="/home/travis/build/LLNL/metis-install" \
        -DUSE_ARPACK=OFF -DCMAKE_BUILD_TYPE=Debug \
        -DSPE10_PERM="/home/travis/build/LLNL/spe10-install/spe_perm.dat" \
        -DSMOOTHG_TEST_PROCS=2 ..;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,3 @@
-#!/bin/sh
 # BHEADER ####################################################################
 #
 # Copyright (c) 2018, Lawrence Livermore National Security, LLC.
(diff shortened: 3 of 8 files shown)


chakshinglee:
The automation in this PR is very nice. Having said that, and I know we have stopped working on "removing dependency on mfem" for a while, but I still think it's reasonable future work. If not too complicated, can we make this automation optional? E.g., set some flag to determine whether to get other dependencies (hypre, metis, etc) from mfem config file? Then there is still other option even if the python script fails to find other dependencies or later on we dont have mfem.

By the way, I was testing it on my tux box, I got the following message, do you know what is the issue? I think `cmake` finds the path to the mfem config file. But something is unexpected in the file.

```
bash-4.2$ ./config/lee_config.sh 
-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Found MPI_C: /usr/apps/mpich2/3.1.4/lib/libmpi.so  
-- Found MPI_CXX: /usr/apps/mpich2/3.1.4/lib/libmpicxx.so;/usr/apps/mpich2/3.1.4/lib/libmpi.so 
Traceback (most recent call last):
  File "parsemk.py", line 140, in <module>
    packages, includes = parse_packages(filename=sys.argv[1], verbose=False)
  File "parsemk.py", line 101, in parse_packages
    name_package(p, expected_names)
  File "parsemk.py", line 52, in name_package
    p["name"] = "unexpected:" + p["libraries"][0]
IndexError: list index out of range
CMake Error at CMakeLists.txt:89 (include):
  include could not find load file:

    out.cmake


-- Found Doxygen: /usr/bin/doxygen (found version "1.8.5") 
-- Configuring incomplete, errors occurred!
```

barker29:
Having this be optional is a good idea, I will do that.

Regarding your error, can you email me a copy of the MFEM config.mk file that it is trying to read?

barker29:
The automation should be optional now. In particular, if you have `-DADOPT_MFEM_DEPENDENCIES=OFF` in your config script, everything else should work just as before with all the dependencies specified explicitly. Let me know how it goes.

chakshinglee:
The new option `-DADOPT_MFEM_DEPENDENCIES=OFF` works like a charm. 

When I turn it `ON`, it seems finding the right things, here is the output if I enforce verbose=True in the python script:
```
packages:

   HYPRE
   rpaths:
   linkpaths:
     /home/lee1029/Git/hypre/install/lib
     /home/lee1029/CASC/metis-install/lib
   libraries:
     HYPRE
     metis

   metis
   rpaths:
     /home/lee1029/CASC/suitesparse-install/lib
   linkpaths:
     /home/lee1029/CASC/suitesparse-install/lib
     /home/lee1029/CASC/metis-install/lib
   libraries:
     klu
     btf
     umfpack
     cholmod
     colamd
     amd
     camd
     ccolamd
     suitesparseconfig
     rt
     metis
     lapack
     blas
```

But when I try to run some examples, it has an error
```
./finitevolume: symbol lookup error: /lib64/libcholmod.so.2: undefined symbol: amd_printf
```
This path is different from the one in `out.cmake`, which is 
```
find_library(cholmod_LIB cholmod PATHS /home/lee1029/CASC/suitesparse-install/lib /home/lee1029/CASC/metis-install/lib )
list(APPEND MK_LIBRARIES ${cholmod_LIB})
```

PR #55, opened 2019-12-13
Add parameter for number of edge traces

chakshinglee:
Add an extra parameter so that we can separately specify maximum number of spectral vertex targets per aggregate and max number of edge traces per coarse face.


src/LocalMixedGraphSpectralTargets.hpp
```diff
@@ -90,6 +92,7 @@ class UpscaleParameters
     UpscaleParameters() : max_levels(2),
         spect_tol(0.001),
         max_evects(4),
+        max_traces(4),
```
barker29, line 95: Is this the right default? If you make it 1 then the tests in `stest.py` don't have to be changed as much.

barker29 (approved):
This looks fine. Travis is failing because the SPE10 dataset has disappeared from the web - I'll try to think of a workaround, maybe just ignoring SPE10 tests for now.

src/LocalMixedGraphSpectralTargets.hpp
chakshinglee: I also thought setting it to be 1 would require less change, but actually in that case 27 tests failed while using 4 as default only failed 14 tests.

PR #56, opened 2020-01-08
Adding FAS example

chakshinglee:
This PR adds abstract nonlinear solver and FAS classes, as well as some concrete example and tests on elliptic problem of the form -div (K(p) grad p) = f, including Picard- and Newton-type linearizations. 

Some tests results are changed, mainly due to two changes - I flipped the x and z directions of SPE10 coefficient (the new one aligns with the original data set), and I added more edge basis to the boundary face. 

There are still two tests failing on Travis. The valgrind one should be fine - it does not show in my tuxbox, and someone on google said it will be gone for newer version of valgrind. Actually can we use newer version on Travis? The `parsamplegraph4` is NNZ problem. NNZ of coarse matrix in my tuxbox, here and my laptop are all different. I think this is due to some very small entries that are supposed to be zero. I do not know how to resolve this, maybe we need to remove the OC as one of the criteria, or use a range. 

In the case of Newton's method, the Jacobian is nonsymmetric, so the solvers are adjusted. In particular, the nonsymmetric part are added, and GMRES will be used in that case. For this reason,  `MinresBlockSolver` is simply renamed to `BlockSolver`. 

P. S. somehow github treats `BlockSolver.cpp` as a new file in the PR diff, even though git knows it was just a rename in the original commit. So for now I keep the `MinresBlockSolver.cpp` file name for review purpose (easier to see the actual modification). I will rename it back to `BlockSolver.cpp` after the review.


barker29:
Is there a hurry to get this merged into master (ie, do you have a collaborator in mind who wants to use this)? I do want to review it, but it's a significant amount of code and I'm not sure I can get to it very soon.

chakshinglee:
It is not in a hurry. I understand this involves a large addition of codes, so please take your time. I make this PR because I am working on FAS for multiphase flow and it's good to tidy up the current code and move on.

chakshinglee:
@francoishamon thanks for agreeing to review this PR. I suggest to focus on three files: `nldarcy.cpp`, `NonlinearSolver.hpp`, and `NonlinearSolver.cpp`. Of course, feel free to comment on other changes, but they (e.g. change in test results) are probably not that relevant for the FAS project.

src/GraphTopology.hpp
```diff
@@ -50,7 +50,7 @@ class GraphTopology
        @param coarsening_factor intended number of vertices in an aggregate
        @return coarse graph
     */
-    Graph Coarsen(const Graph& fine_graph, int coarsening_factor);
+    Graph Coarsen(const Graph& fine_graph, int coarsening_factor, int num_iso_verts = 0);
```
barker29, line 56: Can you document the purpose of the new argument?

src/MatrixUtilities.hpp
```diff
@@ -481,6 +479,10 @@ mfem::SparseMatrix GetDiag(const mfem::HypreParMatrix& mat);
 /// @return "off diagonal block" of a HypreParMatrix
 mfem::SparseMatrix GetOffd(const mfem::HypreParMatrix& mat);
 
+int NNZ(const mfem::SparseMatrix& mat);
```
barker29: I recommend removing this function (why can't the caller just call `mat.NumNonZeroElems()` which is clearer code anyway?)

src/MatrixUtilities.hpp
```diff
@@ -481,6 +479,10 @@ mfem::SparseMatrix GetDiag(const mfem::HypreParMatrix& mat);
...
+
+double FroNorm(const mfem::SparseMatrix& mat);
```
barker29: Let's spell this out `FrobeniusNorm` and add one line of documentation.

src/MetisGraphPartitioner.hpp
```diff
@@ -93,7 +93,9 @@ class MetisGraphPartitioner
        (this does a deep copy of the indices)
     */
     void SetPreIsolateVertices(int index);
-    void SetPreIsolateVertices(const mfem::Array<int>& indices);
+    void SetPreIsolateVertices(std::vector<int> indices);
+    void SetPreIsolateVertices(std::vector<std::vector<int>> sets);
```
barker29: The meaning of this version is pretty clear to me, but maybe not to everyone, so add a quick description to the documentation above (or a separate doc line for this method if you want).

src/MinresBlockSolver.cpp
```diff
@@ -16,32 +16,32 @@
 /**
    @file
 
-   @brief Implements MinresBlockSolver object.
+   @brief Implements BlockSolver object.
```
barker29: Move this file to be `BlockSolver.cpp` (as you already did for `BlockSolver.hpp`)

src/MixedLaplacianSolver.hpp
```diff
@@ -26,6 +26,8 @@
 namespace smoothg
 {
 
+enum KrylovMethod { CG, MINRES, GMRES };
```
barker29: Let's make this an [enum class](https://stackoverflow.com/questions/18335861/why-is-enum-class-preferred-over-plain-enum) and/or maybe move it inside the `MixedLaplacianSolver` class.

src/NonlinearSolver.hpp
```diff
@@ -0,0 +1,166 @@
...
+#define __NONLINEARSOLVER_HPP__
+
+#include "utilities.hpp"
+#include "Hierarchy.hpp"
+
+namespace smoothg
+{
+
+/// Linearization method
+enum Linearization { Newton, Picard };
```
barker29: enum class

src/NonlinearSolver.hpp
```diff
@@ -0,0 +1,166 @@
...
+    double adjusted_tol_;  // max(atol_, rtol_ * || rhs ||)
+    double rhs_norm_;
+    double resid_norm_;
+    double prev_resid_norm_;
+    double linear_tol_;
+
+    NLSolverParameters param_;
+};
+
+enum Cycle { V_CYCLE, FMG };
```
barker29: enum class

src/utilities.hpp
```diff
@@ -233,11 +232,21 @@ void InvRescaleVector(const mfem::Vector& scaling, mfem::Vector& vec);
...
 
+/// @return columns of mat that contains nonzeros
 std::set<unsigned> FindNonZeroColumns(const mfem::SparseMatrix& mat);
 
+/// @return A map such that order of reordered entity align with order of trueentity
 mfem::SparseMatrix EntityReorderMap(const mfem::HypreParMatrix& entity_trueentity,
                                     const mfem::HypreParMatrix& entity_trueentity_entity);
 
+/// Max of absolute values of entries of vec
+double AbsMax(const mfem::Vector& vec, MPI_Comm comm);
```
barker29: Recommend name `ParAbsMax` (it's kind of like `mfem::ParNormlp()`)

src/utilities.hpp
```diff
@@ -233,11 +232,21 @@ void InvRescaleVector(const mfem::Vector& scaling, mfem::Vector& vec);
...
+
+/// Min of entries of vec
+double Min(const mfem::Vector& vec, MPI_Comm comm);
```
barker29: `ParMin` or maybe `ParVectorMin`

examples/nldarcy.cpp
```diff
@@ -0,0 +1,532 @@
...
+ *
+ * smoothG is free software; you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License (as published by the Free
+ * Software Foundation) version 2.1 dated February 1999.
+ *
+ ***********************************************************************EHEADER*/
+
+/**
+   @file nldarcy.cpp
+   @brief nonlinear Darcy's problem.
```
barker29, line 18: We haven't done this before, but it might be nice to have a few example command lines in the comments here (like they do in the MFEM examples). I can run this, but I am not sure what it's doing.

examples/nldarcy.cpp
```diff
@@ -0,0 +1,532 @@
...
+    if (problem == "spe10")
+    {
+        ess_attr[dim - 2] = 0;
+        fv_problem.reset(new SPE10Problem(perm_file, dim, 5, slice, use_metis, ess_attr));
+
+    }
+    else if (problem == "egg")
+    {
+        ess_attr[1] = 0;
+        fv_problem.reset(new EggModel(num_sr, num_pr, ess_attr));
```
barker29, line 213: In general we are missing the files for this. We don't need to package them (like we do for SPE10) but we should at least document where they are coming from.

examples/nldarcy.cpp
```diff
@@ -0,0 +1,532 @@
...
+    }
+    else if (problem == "lognormal")
+    {
+        ess_attr = 0;
+        fv_problem.reset(new LognormalModel(dim, num_sr, num_pr, correlation, ess_attr));
+    }
+    else if (problem == "richard")
+    {
+        ess_attr[0] = 0;
+        fv_problem.reset(new Richards(num_sr, ess_attr));
```
barker29, line 223: I can't get this model problem to run at all.

barker29: ( I get some kind of GMRES breakdown, see below:)
```
Assertion failed: (IsFinite(beta)) is false:
 --> beta = inf
 ... in function: virtual void mfem::GMRESSolver::Mult(const mfem::Vector&, mfem::Vector&) const
 ... at line 570 of file: linalg/solvers.cpp
```

examples/nldarcy.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    EllipticFAS nlmg(hierarchy, kappa, ess_attr, mg_param);
+    nlmg.SetPrintLevel(1);
+    nlmg.SetRelTol(1e-6);
+    nlmg.SetMaxIter(200);
+    nlmg.Solve(rhs, sol_nlmg);
+
+    serialize["nonlinear-iterations"] = picojson::value((double)nlmg.GetNumIterations());
+
+    if (visualization)
```
barker29, line 262: I cannot get any visualization to work, I get the cryptic error:
```
Vector::Add(const double, const Vector &)
```

examples/pde.hpp
```diff
@@ -258,44 +242,46 @@ void InversePermeabilityCoefficient::InversePermeability(const mfem::Vector& x,
...
-double InversePermeabilityCoefficient::InvNorm2(const mfem::Vector& x)
+double InversePermeabilityCoefficient::FroNorm(const mfem::Vector& x)
 {
     mfem::Vector val(3);
     InversePermeability(x, val);
-    return 1.0 / val.Norml2();
+
+    for (int i = 0; i < val.Size(); ++i)
+    {
+        val[i] = 1.0 / val[i];
```
barker29, line 300: In 2D case won't this divide by zero? (`val.Size()` is always 3)

barker29 (changes requested):
Overall this looks very good, there are a lot of good simplifications/improvements and the nonlinear solvers will be very useful.

The main thing I want to see before merging is for the `nldarcy` example to run with a visualization I can understand. I don't expect to understand everything you've done here but I would like this example to be clear and to make sense to me.

PR #57, opened 2020-03-27
Upstream changes

barker29:
This PR proposes contributing some changes that were used to support another project into the main github smoothG repo.

At the moment several tests are failing, which probably needs to be resolved before this can be merged.

### DIFF
--- a/examples/pde.hpp
+++ b/examples/pde.hpp
@@ -92,6 +92,19 @@ public:
                                    SliceOrientation orientation = NONE,
                                    int slice = -1);
 
+    /**
+       This fakes a field that is uniform but anistoropic, with given
+       horizontal and vertical permeabilities.
+
+       (the typical use case they are more like *covariances* than
+       permeabilities, but this class doesn't know about that.
+    */
+    InversePermeabilityCoefficient(MPI_Comm comm,
+                                   const mfem::Array<int>& N,
+                                   const mfem::Vector& h,
+                                   double horizontal_perm,
+                                   double vertical_perm);
+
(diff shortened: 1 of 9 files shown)


chakshinglee:
The bug is most likely due to the changes I made in `SparseIdentity` and `SparseDiag`. Specifically, there is some issue when the size is 0. I have made a commit to take care of this special case, please check if it works.

There are other tests that are failing. I think this is because of the way we define `unit_weight` is different. The current definition should be more appropriate, so we should update the test results accordingly.

examples/pde.hpp
```diff
@@ -861,12 +981,15 @@ SPE10Problem::SPE10Problem(const char* permFile, int nDimensions, int spe10_scal
     }
 
     InitGraph();
-    ComputeGraphWeight(unit_weight);
+    // TODO FIXME: unit_weight is the wrong name for what I am doing here
```
chakshinglee: If we want to return edge weights corresponding to the case of constant 1 permeability, then I think the updated approach in this PR is doing the right thing. Note that in this case the edge weight is not 1, but weighted by geometrical quantities like face area, cell center distance, etc.

In the original examples (`sampler`, `ml-sampler`, `ml-sampler4`), we used a pure graph with edge weights being 1 (geometrical quantities are not considered). So we need to update the results if we want to honor the finite volume discretization.

barker29: Based on these comments I removed the option, and rebaselined the `sampler` tests. Now all the tests pass. If you're happy with this and think the rebaselining was legitimate, then I think we're all set.

chakshinglee (approved):
Your new changes look good to me, and all the test pass now.

